PR #33, opened 2021-10-12
Fig missing % in .gdbinit

Sparrrgh:
There is a % symbol missing in .gdbinit. This makes "legend, regs" invisible


ConnorNelson:
Thanks!

PR #60, opened 2022-08-04
install expect (specifically for unbuffer)

ReubenBeeler:
unbuffer is a simple tool used for unbuffering piped messages, so a process can read a challenge output, parse it, and respond accordingly before the challenge terminates.


zardus:
Thanks!

ReubenBeeler:
The process interpreter for unbuffer is `tclsh8.6`, but the challenge requires `sh` or `bash` lol. Thanks for the flexibility

PR #61, opened 2022-08-06
Adding a Desktop tab with novnc and a (minimal) xfce environment

zardus:
This adds novnc in a navbar tab (not a vscode pane, though we can add that easily). The desktop environment is xfce4, because gnome-session seems to depend on a system-wide dbus running.


zardus:
Okay, unless my last push screwed something up, this is ready to go! I've trimmed down XFCE4 as much as I can, and we are ready to roll!

The desktop:

Screenshot from 2022-08-09 00-44-43

angr-management in the browser: 

Screenshot from 2022-08-09 00-44-27

zardus:
Minor consideration: maybe we should pre-insert `/challenges` into the Qt and GTK lists of favorite locations. The QT one is in `~/.config/QtProject.conf`, and the GTK one might be in `~/.config/gtk-3.0/bookmarks`, but making these files in the entrypoint script is a pain in the ass.

zardus:
Neat feature of xfce: it can save a session, and open up the configuration of applications at next launch. We could have this pop up for every challenge:

image

PR #71, opened 2022-08-16
Add support for admins to drop in to monitor and control user desktops

zardus:
Should really help with student support!


ConnorNelson:
Lets:
- extract the docker logic into a util function `get_active_users`
- extract workspace/desktop socket redirect into a util function `redirect_user_socket`
- remove `view_specific_desktop`, set default `user_id` in `forward_desktop` to None (which is checked for to become `get_current_user().id`)

PR #77, opened 2022-08-18
Add shared directories, one for global sharing and one for admin->user sharing.

zardus:
This is more useful for private dojo deploys than the public pwn.college instance.


zardus:
Should be squashed when we merge this.

PR #94, opened 2022-09-21
Add networking to user containers with firewall rules

ConnorNelson:
```
docker network create --opt com.docker.network.bridge.name=user_firewall user_firewall

iptables -I DOCKER-USER -i user_firewall -j DROP
iptables -I DOCKER-USER -i user_firewall -d 1.1.1.1 -j ACCEPT

docker run -it --rm --net=user_firewall pwncollege_challenge
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,8 @@ services:
   challenge:
     build: challenge
     entrypoint: /bin/true
+    networks:
+      - user_firewall
 
   ctfd:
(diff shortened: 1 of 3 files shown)


ConnorNelson:
TBD: add "--force-recreate" to https://github.com/pwncollege/dojo/blob/master/etc/systemd/system/pwn.college.service#L10
to prevent errors?

PR #99, opened 2022-10-04
Listen on unix socket in nginx

ConnorNelson:


ConnorNelson:
This has clearly stalled out, and this definitely isn't the way to do this. When we implement this, it's going to be over TCP.

PR #119, opened 2022-12-13
Wip/git dojos

zardus:


dojo_plugin/api/v1/dojo.py
```diff
@@ -0,0 +1,244 @@
...
+                N=b"\n"
+                assert returncode == 0, (
+                    f"Dojo clone failed with error code {returncode}:<br>"
+                    f"<code>{output.replace(N,b'<br>').decode('latin1')}</code><br>"
+                    "Please make sure that you properly added the deploy key to the repository settings, "
+                    "and properly entered the repository URL."
+                )
+
+                # figure out the dojo ID
+                dojo_specs = list(clone_dir.glob("*.yml"))
```
ConnorNelson, line 129: Maybe we shouldn't offer such flexibility. Let's just check for `dojo.yml`. In the future we might want other ymls.

dojo_plugin/api/v1/dojo.py
```diff
@@ -0,0 +1,244 @@
...
+                )
+                dojo_id = dojo_specs[0].stem
+                assert id_regex(dojo_id), (
+                    f"Your dojo ID (the extensionless name of your .yml file) must be a valid URL component."
+                )
+
+                # make sure there aren't any symlink shenanigans
+                assert dojo_specs[0].is_file() and not dojo_specs[0].is_symlink(), (
+                    f"{dojo_specs[0].name} is not a regular file!"
+                )
```
ConnorNelson, line 141: We can just open `dojo.yml` with O_NOFOLLOW

dojo_plugin/api/v1/dojo.py
```diff
@@ -0,0 +1,244 @@
...
+
+                # make sure we're not overwriting unintentionally
+                dojo_permanent_dir = pathlib.Path(DOJOS_DIR)/str(user.id)/dojo_id
+                if data.get("dojo_replace", False) not in ("true", True, 1):
+                    assert not dojo_permanent_dir.exists(), (
+                        f"You already have a cloned dojo repository containing a dojo with ID {dojo_id}."
+                    )
+
```
ConnorNelson, line 149: We could clone straight to where we are going without needing to do a tmp version.
Git will yell loud if the destination directory isn't empty.

Add a dojo-update capability to change the git remote.

dojo_plugin/api/v1/dojo.py
```diff
@@ -0,0 +1,244 @@
...
+                # make sure we're not overwriting someone else's dojo
+                existing_dojo = Dojos.query.filter_by(id=dojo_id).first()
+                if existing_dojo is not None:
+                    assert existing_dojo.owner_id == user.id, (
+                        f"A dojo with the ID {dojo_id} was already created by a different user. Please choose a different ID."
+                    )
+                    join_code = existing_dojo.join_code
+                else:
+                    join_code = random_dojo_join_code()
+
```
ConnorNelson, line 159: Maybe we really should just not let people have the potential for name collisions on dojos.

Especially for private dojos, it's confusing to not be able to see a private dojo, and then have to just guess names until you find an available one.

We make urls for non-official dojos be something like:
`/<chosen_name>-<owner_user_id>/`

Being an official dojo can mean having a NULL owner id, getting a "prettier" url, but still keeping admin/whatever rights on your dojo.

dojo_plugin/api/v1/dojo.py
```diff
@@ -0,0 +1,244 @@
...
+                if dojo_permanent_dir.exists():
+                    shutil.rmtree(dojo_permanent_dir)
+                dojo_permanent_dir.parent.mkdir(parents=True, exist_ok=True)
+                shutil.move(clone_dir, dojo_permanent_dir)
+
+                # load it for real!
+                log_handler.reset()
+                load_dojo(
+                    dojo_id, (dojo_permanent_dir/(dojo_id+".yml")).read_text(),
+                    user=user, commit=True, dojo_dir=dojo_permanent_dir, log=logger, initial_join_code=join_code
```
ConnorNelson, line 184: We should make `join_code`s specified in the dojo yml.
We should make public vs private specified in the dojo yml (default to private).

ConnorNelson (commented):
Bunch of random thoughts while reviewing

dojo_plugin/utils.py
```diff
@@ -408,3 +386,67 @@ def belt_challenges():
         )
         for color, categories in color_categories.items()
     }
+
+# based on https://stackoverflow.com/questions/36408496/python-logging-handler-to-append-to-list
+class ListHandler(logging.Handler): # Inherit from logging.Handler
```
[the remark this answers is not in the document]
zardus: ah, it's an orphan of a refactor

dojo_plugin/utils.py
```diff
@@ -408,3 +386,67 @@ def belt_challenges():
...
+    def __init__(self, log_list):
+        logging.Handler.__init__(self)
+        self.log_list = log_list
+
+    def emit(self, record):
+        self.log_list.append(record.levelname + ": " + record.getMessage())
+
+class HTMLHandler(logging.Handler): # Inherit from logging.Handler
```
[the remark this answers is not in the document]
zardus: hahaha, indeed. That might have been ChatGPT leftovers? No idea how that got there :-)

dojo_plugin/api/v1/dojo.py
zardus: There is no other way to specify dojo IDs right now, so we either DOJO_ID.yml or get the ID somehow from the yml (whether it's ID-ifying the dojo name or having an ID field there).

In general, there are probably massive issues if the dojo ID changes in an update, btw. Challenges will be unable to be found and will get recreated, with no solves. I don't check for that, but it's something to keep in mind.

dojo_plugin/api/v1/dojo.py
zardus: Either way works for me, but then the pathlib integration isn't as magic...

dojo_plugin/api/v1/dojo.py
zardus: We use cloning to update, so we can't just mandate an empty dir. If we want to avoid using clones for updates, we need to make sure that `git pull --force` doesn't have any corner cases where things can go wrong. I'm okay either way, ofc; it'd make things a lot simpler in the dojo loading, because we won't need to pass a directory path around.

dojo_plugin/api/v1/dojo.py
zardus: I'm not against that route (or `pwn.college/dojos/UID/DOJO_ID`). We could have a completely different, admin-maintained functionality to promote a dojo to the global namespace (`pwn.college/dojos/GLOBAL_DOJO_ID`) via another unique field in the dojos table.

dojo_plugin/api/v1/dojo.py
zardus: I like this idea (as long as we do the necessary check here for public dojos).

dojo_plugin/utils.py
```diff
@@ -2,41 +2,60 @@
...
-from sqlalchemy import String, DateTime, Integer
-from sqlalchemy.sql import and_, or_
+from CTFd.utils.security.sanitize import sanitize_html
+from sqlalchemy import String, Integer
+from sqlalchemy.sql import or_
 
 
 CHALLENGES_DIR = pathlib.Path("/var/challenges")
 DOJOS_DIR = pathlib.Path("/var/dojos")
+HOST_DOJOS_DIR = pathlib.Path("/opt/pwn.college/data/dojos")
```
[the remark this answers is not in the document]
zardus: I'll leave all this path specification stuff up to your discretion :-)

PR #153, opened 2023-03-31
Add docker registry

ConnorNelson:


ConnorNelson:
This has clearly stalled out. Maybe it's a feature we will still want, and in fact, #397 is likely to involve some sort of registry in the infrastructure, but we'll consider this later, maybe (after #397).

PR #158, opened 2023-04-06
Allow KPTI to be disabled for vm challenges

zolutal:
KPTI is the Linux kernel's mitigation for meltdown, support configuring it for vm challenges in a similar way to KASLR to make meltdown challenges possible.


robwaz:
:+1:

PR #159, opened 2023-04-07
Update outdated README

Scoder12:
Should make setting up a development environment easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,43 +16,58 @@ The associated challenge binary may be either global, which means all users will
 ## Setup
 
 Clone the repository:
+
 ```sh
 git clone https://github.com/pwncollege/dojo /opt/dojo
 ```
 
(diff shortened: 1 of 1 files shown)


ConnorNelson:
Super minor, but I prefer to put each sentence on its own line. This makes it easier when reading diffs to immediately see what sentences have been changed.

Also, as you said in your container-setup.sh PR, we should update the language now to reflect that `challenge-minimal` is now the default.

Scoder12:
Zero idea why the diff looks like that.

Scoder12:
Fixed it (hackily) with a force push. I don't think I rebased properly or something.

ConnorNelson:
You had diff data in there (e.g. `<<<<<<< HEAD`). I went ahead and cleared that out, otherwise looks good to me!

Issue #70, opened 2022-08-16
Move /opt/pwn.college to /opt/dojo inside infrastructure


ConnorNelson:
Stick to the dojo branding.

PR #160, opened 2023-04-07
Update container-setup.sh

Scoder12:
- paramaterize dojo dir (progress towards #70)
- remove reference to outdated data_example


Scoder12:
TODO: Update README section in #159 on challenge-minimal now that it is the default.

PR #179, opened 2023-05-24
add a non-slash version of the dojo url

zardus:
Nicer this way, but maybe breaks stuff?


ConnorNelson:
I would be fine with a 302 redirect. Design wise, the idea is that a dojo has several resources underneath it (at a minimum for now, the modules), and this is the root resource, and so it should end with a slash, since it is a container.

zardus:
I went rogue and did this, but maybe a 302 would be better

PR #182, opened 2023-05-31
Navbar login link populates "next"

### LINKED ISSUE #102: QOL: Make login in header always redirect back to previous page
Right now, JS redirects to the previous page but the login link in the header does not.

Scoder12:
Now matches the behavior of JS. When pressing "login" on a challenge page, you will be redirected back after logging in!

Closes #102


ConnorNelson:
Great, thanks!

Issue #73, opened 2022-08-16
Support other edu emails


ConnorNelson:
Rather than checking if an email ends with `.edu`, we should check if it ends with `\.edu(\.\w+)?`

zardus:
Relevant: https://github.com/Hipo/university-domains-list

PR #185, opened 2023-07-10
Change student icon check in scoreboard.py

Zeeshan12340:
If an email has the format: `<name>@student.edu.<country>` or any other format where `.edu` is not at the end, then the email is not recognized as a student email. Probably broad but this is a simple fix.


ConnorNelson:
Thanks for the PR! I agree that this is overly broad, but it's better than the current.

Closes #73

PR #188, opened 2023-07-17
Add belt awarding bot command

robwaz:
Working draft of bot awarding belts.  Currently assigned to command `/check_belts` (sensei only).  It currently only prints out belt and username.  Actual role modification line is commented out to prevent unintended shenanigans (Line 305 bot.py).

Possible improvements:
- Have SQL query determine 100% completion
- Don't use pandas.. I like it ¯\_(ツ)_/¯
- Turn into a daily run instead of a command, command is just for POC

### DIFF
--- a/discord_bot/Dockerfile
+++ b/discord_bot/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.9-slim-buster
 
-RUN pip install discord-py
+RUN pip install discord-py sqlalchemy pandas PyMySQL
 
 ADD bot.py /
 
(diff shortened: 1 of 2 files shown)


robwaz:
@ConnorNelson could you sanity check this before I inadvertently make everyone a black belt?

ConnorNelson:
I have no clue how this passed by my radar, but YOLO.

I think in the long term we'll want to do belt assignment checks on flag submission. But I'm super stoked to have some version of automated belting back. Thanks!

PR #190, opened 2023-07-20
dojo share button 

Zeeshan12340:
Currently, there are no references on how to share your dojo using the dojo invite link on the website itself. This pull request gives  a way to share your dojo invite link, clicking on the share icon right next to the admin icon in your dojo copies the link to your clipboard. This is only available to the dojo admin. 

Image of the addition and what happens on clicking.
image

PR #191, opened 2023-07-20
Add button to share dojo

ConnorNelson:
I couldn't edit the PR in #190, so here we are. This closes #190.


ConnorNelson:
Overall, I like the idea, just refactored it slightly. Thanks!

PR #194, opened 2023-08-06
Minor challenge container QoL improvements

zardus:
The docker-entrypoint.sh for the challenge container was getting cumbersome, so I split it out to into pieces. Also some other minor stuff, the biggest being an INSTALL_IDA parameter to configure including or not including IDA Free (for situations where IDA is not allowable).


zardus:
Life is so easy without CI

zardus:
Unsurprisingly, this broke everything. Rolling back...

Issue #152, opened 2023-03-31
Fix Nested Nav Bar


ConnorNelson:
It happens from the workspace/desktop socket failing to be accessible. This produces a 404, in the iframe, which includes the bar. We need to detect the 404 on Unix socket redirect, and do something smarter, probably make it not possible to fail, but show something better if it does.

Scoder12:
Simple way to repro a version of this: 
```diff
diff --git a/dojo_plugin/pages/workspace.py b/dojo_plugin/pages/workspace.py
index 2174231..f8ba8f0 100644
--- a/dojo_plugin/pages/workspace.py
+++ b/dojo_plugin/pages/workspace.py
@@ -22,6 +22,7 @@ def view_workspace():
 @workspace.route("/workspace/<path:path>")
 @authed_only
 def forward_workspace(path=""):
+    return redirect("/")
     prefix = "/workspace/"
     assert request.full_path.startswith(prefix)
     path = request.full_path[len(prefix):]
 ```

ConnorNelson:
My understanding (and after briefly testing it), vscode/desktop can no longer throw 404s after our move from unix to tcp. So, this is no longer relevant.

PR #204, opened 2023-08-17
detect iframe redirects and redirect top window

Scoder12:


ConnorNelson:
Can you help me understand what this PR addresses, and what its effect is? It seems like it maybe addresses #152?

Scoder12:
Yep, exactly. It's meant to address #152. As far as I understand it, #152 is caused by the having being HTTP redirected to the login page. This PR detects whether the iframe was HTTP redirected before loading by checking it's `location.href` before and after `onload` fires. If the URLs are different, it redirects the top frame to the new one.

zardus:
@ConnorNelson are we good to merge this?

Issue #126, opened 2023-01-19
Add CI


adamdoupe:
Would be great to have some testing of `dojo.sh`

PR #208, opened 2023-09-01
Implement a CI system. Initial test.

adamdoupe:


zardus:
Whoa no way!!! Is this for real?

adamdoupe:
Ok, this is now good to go. 

Fixes #126.

Right now the builds use too much disk space for GitHub actions caching to help: probably we will want to run our own self-hosted runner in the future (or pay $ for GitHub Teams and get bigger runners).

Also, this currently ensures:

- The dojo creation works
- The inner dojo creates the website enough to get a `200` back. It uses the `test/test_running.py` script. We should extend this in the future to add more.

Note for the future: the dojo has to be run with port 2222 exposed instead of 22, because the runners already claim port 22.

PR #209, opened 2023-09-02
support disabling of dojo joining

zardus:
This adds a `joinable` tag to dojos, which allows the disabling of new joins to a dojo.

### DIFF
--- a/dojo_plugin/models/__init__.py
+++ b/dojo_plugin/models/__init__.py
@@ -60,6 +60,7 @@ class Dojos(db.Model):
     description = db.Column(db.Text)
 
     official = db.Column(db.Boolean, index=True)
+    joinable = db.Column(db.Boolean)
     password = db.Column(db.String(128))
 
     data = db.Column(db.JSON)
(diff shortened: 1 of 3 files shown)


ConnorNelson:
What is the usecase for someone not being able to join a dojo? Official dojos are already not joinable (except as a student for a course dojo), I don't think we even really need to keep requiring community dojos to be joinable, they should just follow the same model as official dojos. Finally there are private dojos, which have a password to gatekeep joining.

zardus:
But it passed CI.....

Honestly, I'd forgotten about passwords. Do we have an example of a password-protected dojo I could play around with?

adamdoupe:
@zardus CI is only testing that both parts build and that the website comes up. We need to actually add test cases that exercise functionality so you can get the warm fuzzy of CI
https://github.com/pwncollege/dojo/pull/208#issuecomment-1703698912

ConnorNelson:
It may have passed Continuous Integration, but it hasn't passed Connor Integration 😉

There aren't any open source private dojos, since that wouldn't make any sense (maybe the password shouldn't be specified in the yml?).

https://github.com/pwncollege/cse365-2023-summer/blob/main/dojo.yml is an example of a dojo with a password. It's just `password: MYPASSWORD` at the top level. I probably wouldn't modify that dojo though...

For any on lookers, don't worry, there aren't any fun private challenges in that dojo.

PR #214, opened 2023-09-04
ssh should not ask for password

Zeeshan12340:
If someone has not configured ssh or is using an incorrect key, the dojo asks for a password which is very confusing. `PasswordAuthentication no` is set in sshd_config but it needs an extra setting `ChallengeResponseAuthentication no` to deny password auth completely.
https://blog.tankywoo.com/2013/09/14/ssh-passwordauthentication-vs-challengeresponseauthentication.html


ConnorNelson:
Merged and tested, it works great! Thanks @Zeeshan12340!

Issue #184, opened 2023-06-30
Add wechall.net API to dojo for even more internet points


goreil:
# Why? Because it's awesome.
[wechall.net](www.wechall.net) is a global ranking list of multiple wargames like OverTheWire, pwnable.kr and Cryptohack.
By gathering points on the respective websites, users get points on a global ranking ladder! By connecting pwn.college to WeChall, there is even more incentive to get flags, since bragging rights are a big motivation (at least to me). It also gives us more exposure on a big wargame forum.

## Steps
The simple procedure is as follows:
1. Implement an API that validates an user. 
2. API that returns the score for a user
3. Contact the admins of [wechall.net](www.wechall.net)

If scores are public then AUTHKEYS are not needed in the APIs.
The API's must be accessible via GET request.

### Step 1 Implement an API that validates an user. 
Example:
`pwn.college/validate?username=%USERNAME%&email=%EMAIL%[&authkey=%AUTHKEY%]`
This script must return simply "1" OR "0",
1:email/username combination exists.
0:combination does not exist or authkey wrong.

### Step 2 API that returns the score for a user
Example:
`pwn.college/score?username=%USERNAME%[&authkey=%AUTHKEY%]`
Your output must contain at least userscore and maxscore. So the output of this script could be like `userscore:maxscore`
Perfect output for this script is: `username:rank:score:maxscore:challssolved:challcount:usercount`

### Step 3 Contact the admins
When the 2 API's are implemented, we can write a mail to the admin.

Email: gizmore@wechall.net
First Name: Christian
>I also want to point out that every single site is a great addition to the challenge scene. Shouts and thanks go out to all the creators! - wechall.net

The detailed requirements can be found on [https://www.wechall.net/join_us](https://www.wechall.net/join_us)

gizmore:
The scoring API should support all supported fields to have the best user and site admin experience.
Like we can track usercount and challcount and support different scores for different challenges.
Of course we do not share our jpgraphs with 3rd parties :P ....

The optimal userscore api format is 6 fields:

                # rank:score:maxscore:challsolved:challcount:usercount
		list($rank, $score, $maxscore, $challsolved, $challcount, $usercount) = explode(':', $result);

For pwn.college, we could abuse the challenge count fields to reflect belts or courses.
Also, new general "trophy" functions would be cool and imaginable, like a PwnC yellow belt trophy :)

greez
giz

Zeeshan12340:
[...]

PR #217, opened 2023-09-12
Wechall/additional

Zeeshan12340:
basically, adding the extra stuff for wechall.net #184 , global rank and user count, `global_scoreboard_data` is the same as `dojo_scoreboard_data` with arguments ignored and duration filter query removed so maybe it's not even needed. Other things worked fine for me and i tested for some edge cases but i obviously don't have the actual pwn.college dataset.

P.S: I don't know why these previous commits are still here.


ConnorNelson:
So `ignore_visibility` actually has to do with visibility start/stop times. What we really want to do if we want official only, is `.filter_by(official=1)`. I'm also not completely sure what's going to happen when two official dojos share a challenge, but *probably* it will result in duplicated solves in this query. I think we need some sort of `.distinct(Solves.challenge_id)` or something like that before we group.

Zeeshan12340:
Whoops, completely misread what `ignore_visibility` was doing. `DojoChallenges.solves().filter(Dojos.official == 1).distinct(Solves.challenge_id)`  seems like it filters for the official dojos correctly. I'll have to test the `distinct ` query because I do not have dojos that share challenges setup.

dojo_plugin/api/v1/score.py
```diff
@@ -44,4 +59,9 @@ def get(self):
         user_score = sum(1 for solve in user.solves if solve.challenge_id in official_challenge_ids)
         max_score = len(official_challenge_ids)
 
-        return f"{user_score}:{max_score}"
+        user_count = len(Users.query.all())
```
ConnorNelson, line 62: Can we do `Users.query.count()`

```
In [77]: %time Users.query.count()
CPU times: user 4.96 ms, sys: 66 µs, total: 5.02 ms
Wall time: 18.7 ms
Out[77]: 38869

In [78]: %time len(Users.query.all())
CPU times: user 5.13 s, sys: 281 ms, total: 5.41 s
Wall time: 5.42 s
Out[78]: 38869
```

ConnorNelson (changes requested):
Happy to merge with this simple change. Thanks!

dojo_plugin/api/v1/score.py
Zeeshan12340: Done!

PR #219, opened 2023-09-14
WIP: Initial implementation of daily backups to cloud storage. 

adamdoupe:


ConnorNelson:
Also, I think it's totally fine, and maybe even preferable, if we write all of the backup logic in python, and throw it in as a docker service.

If we're concerned about reliability, we can include a heartbeat discord ping that keeps us in the loop that our backups are firing every day or every week or however often we want a heartbeat discord ping.

PR #221, opened 2023-09-18
Windows!

Scoder12:
- Build windows VM gated behind `WINDOWS_VM` config variable
- Add `/desktop-win` route
- Add `windows` script to challenge container


windows/Dockerfile
```diff
@@ -0,0 +1,87 @@
+ARG UBUNTU_VERSION=20.04
```
ConnorNelson, line 1: Does this need to be parameterized?

windows/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+  && apt-get update \
+  && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    ca-certificates apt-transport-https curl \
+    xorriso mtools dosfstools qemu-utils qemu-kvm \
+  && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# IDA Freeware: only permissible for free and open deployments of the dojo!
+ARG INSTALL_IDA=False
```
ConnorNelson, line 16: No IDA

windows/full.yml
```diff
@@ -0,0 +1,16 @@
+services:
+  windows:
+    build:
+      context: ./
+      target: ${WINDOWS_VM}
+      args:
```
ConnorNelson, line 6: No args

challenge/docker-entrypoint.d/10_start_vm.sh
```diff
@@ -1,3 +1,14 @@
...
 
-find /challenge -name '*.ko' -exec false {} + || vm start
+if ! find /challenge -name '*.ko' -exec false {} +
+then
+  vm start
+else
+  chmod -s "$(which vm)"
+fi
+
+if [[ -f /challenge/challenge.exe ]]; then
```
ConnorNelson, line 10: Can we do the same '*.exe' style, to allow more flexibility in challenge names?

Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN find /opt/pwn.college/script -type f -exec ln -s {} /usr/bin/ \;
 EXPOSE 22
 EXPOSE 80
 EXPOSE 443
+EXPOSE 5912
```
ConnorNelson, line 44: I don't think we need this

challenge/docker-entrypoint.d/39_start_gui.sh
```diff
@@ -34,6 +34,21 @@ start-stop-daemon --start \
                   </dev/null \
                   >>/tmp/vnc/websockify.log \
                   2>&1
+
```
ConnorNelson, line 37: Can we refactor this so that we have a: `38_start_gui_windows.sh`

All of our pre-existing stuff stays here, all of the new windows stuff moves there. Additionally, instead of "2" suffix in all of these paths, lets use "-windows".

ConnorNelson: Additionally, we should do the same "*.exe" check to determine if this should be started.

windows.md
```diff
@@ -0,0 +1,131 @@
+# Dojo Windows Infrastructure
```
ConnorNelson, line 1: Can we move this file into windows/ as as `README.md`

windows/none.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  windows:
+    build:
+      context: ./
+      target: none
+      args:
```
ConnorNelson, line 6: No args

ConnorNelson (changes requested):
This is obviously super amazing! Lets make just these few minor changes, and I'll go ahead and merge. We'll get around to some of the other refactor ideas we discussed later, but I'm excited to get this merged ASAP!

ConnorNelson:
Also, is it possible to include the challenge-proxy.exe source code, and for us to build it during the build process?

windows/Dockerfile
Scoder12: It is in the challenge image. If we match that ubuntu version, you don't have to pull two different ubuntu images.

windows/Dockerfile
Scoder12: As in, remove IDA since the linux one is sufficient?

challenge/docker-entrypoint.d/10_start_vm.sh
Scoder12: Even though `ChallengeProxy` can only run `challenge.exe`?

Dockerfile
Scoder12: > The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. 
This is simply to document that if you want to see the output of the build process, you should forward this port to be able to connect in with VNC.

Scoder12:
> Also, is it possible to include the challenge-proxy.exe source code, and for us to build it during the build process?

That would make the workflow for updating it a lot easier, so I'll do it. The tradeoffs are: it requires pulling in the rust compiler (which takes some disk space), and I'm concerned that if the source code gets complex enough, the rust compiler could use up all the memory (although I think the current 4G is enough to build the source code as it stands today).

ConnorNelson:
I see. Yeah I would say its worth it for maintainability (if we need to change the proxy in the future, for example).

ConnorNelson:
Also, are these are only big (slow) downloads:
- https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/latest-virtio/virtio-win.iso
- https://software-static.download.prss.microsoft.com/sg/download/888969d5-f34g-4e03-ac9d-1f9786c66749/SERVER_EVAL_x64FRE_en-us.iso

I think I'm going to also look into refactoring this to use a build cache so we don't have to redownload each build.

windows/Dockerfile
ConnorNelson: This image is just used for building windows though, right? I think the reason we parameterized it on the challenge image is to support difference ubuntu bases for challenge runtimes (e.g. a 20.04 libc vs a 22.04 libc). If this is just a build helper, I don't think it makes sense to parameterize it.

windows/Dockerfile
ConnorNelson: Oh right, I forgot you install IDA on windows as well. Disregard!

Dockerfile
ConnorNelson: I see. Are we concerned about the build process failing? Obviously this makes sense from a dev perspective, but I think I see these EXPOSEs here as an interface to dojo-consumers. Our black-box dojo speaks on 22/80/443 for users.

ConnorNelson (changes requested):

ConnorNelson:
> Also, are these are only big (slow) downloads:
> 
> * https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/latest-virtio/virtio-win.iso
> * https://software-static.download.prss.microsoft.com/sg/download/888969d5-f34g-4e03-ac9d-1f9786c66749/SERVER_EVAL_x64FRE_en-us.iso
> 
> I think I'm going to also look into refactoring this to use a build cache so we don't have to redownload each build.

Nevermind! I see you're already caching it because these are run-time downloads. I understand now.

windows/Dockerfile
Scoder12: True, I don't want bumping the challenge ubuntu version to mysteriously break the windows builder either.

ConnorNelson:
Documenting here, it looks like the windows command depends on `psutils`, which is not in the default minimal challenge container.

Scoder12:
OK to add that as a dependency?

Scoder12:
I could also probably do this just by iterating over `/proc/*/exe` manually if that's better.

ConnorNelson:
I'm realizing now, you're probably assuming the full challenge-image, which is probably a fairly reasonable expectation, considering the linux vm also depends on it (for qemu-system-x86). I am currently taking the minimal image and determining what is required to get this thing booted. For now it's probably fine if we expect the full challenge image, but maybe theres something we can do in the future to just get whats necessary for Linux/Windows VMs.

ConnorNelson:
We might want to remove the psutil dependency, that way we can just install `qemu-system-x86` and maybe also `qemu-utils`.

ConnorNelson:
Also for reference, how long did it take you to build the Windows image? I think it just took the production server an hour. Can windows take advance of more cores/memory to build faster?

Scoder12:
Odd, that seems like a lot more than it takes on my machine. There's a few phases, the first one being copying the ISO which is definitely disk-IO bound, and then installing software (visual studio build tools takes most of the time), which I *think* is also mostly downloading and extracting big files. We can up the CPUs and memory allocated, it's just hardcoded to 4 threads and 4096M right now.

ConnorNelson:
OK, it looks like we also need `builder-virtiofsd` (which makes sense).

I'm envisioning like a `challenge-vm`, `challenge-vm-linux`, and `challenge-vm-windows`. The linux one builds the kernel, the windows one builds `virtiofsd`, and eventually, windows itself.

ConnorNelson:
Yeah, I might try that. Server CPUs are definitely much slower as well. There's a lot of them, but any individual one isn't so good. Either way, taking an hour to build isn't the end of the world though. Especially with good cache systems that make it so we only really have to build windows once, and then whenever we update it.

ConnorNelson:
> OK, it looks like we also need `builder-virtiofsd` (which makes sense).
> 
> I'm envisioning like a `challenge-vm`, `challenge-vm-linux`, and `challenge-vm-windows`. The linux one builds the kernel, the windows one builds `virtiofsd`, and eventually, windows itself.

And since it looks like I'm doing a live update of dependencies from minimal at this point. We also need `ssh`.

ConnorNelson:
And with that, I've officially accessed the windows VM from the production server! Amazing!

Scoder12:
Do you think there's any way to put the windows script later in the build process? Right now it invalidates a ton of layers which makes changing it super annoying to test.

ConnorNelson:
Is there anything we can do about not needing to enter a password of `hacker` with `vm connect`? I think even automating the password enter might be more ideal if we can't make it password-less. Or somehow synchronizing an ssh key.

PR #222, opened 2023-09-18
Use the correct executable name for suid python3.

ltfish:
`/usr/bin/python` does not exist in the minimal build of challenge containers.


ConnorNelson:
Thanks!

PR #252, opened 2023-11-13
Fuzz Harnessing Dojo

swirsz:


ConnorNelson:
This would fit much better as its own dojo, like https://github.com/pwncollege/example-dojo. Probably locate it at somewhere like `sefcom/fuzz-dojo`.

The reason for that, is that this is a *massive* proposed change, and it would make sense to decouple the overall infrastructure from this particular challenge set's dependencies. That way when we, for example, upgrade to a base image of 22.04 or 24.04, your dependencies aren't hit in the crossfire. This will make maintenance easier for `pwncollege/dojo` and for your `sefcom/fuzz-dojo`.

https://github.com/pwncollege/example-dojo shows how you specify custom dojos with their own challenges (take a look through example-dojo to better understand the structure). This does not show you how to specify a custom challenge image/Dockerfile, but we are currently working on making that work.

swirsz:
If it is running under its own dojo:

  1) would it be possible to gate the dojo behind an IRB acceptance form that you must accept to continue?
  
  2) either share or synchronize the user database?

If users 100% always had the same account on both dojos, we could still run reports to evaluate user performance on the main dojo versus user performance on the "fuzz" dojo.  IE: be able to say that users who did well in pwn-college module "X" also performed well in the fuzz driver module.

Synchronization could also just be one way to avoid tombstoning and other issues: force users to always register on the main server just by redirecting the register button to the main server?

p.s. My test setup is an independent dojo currently. I made a production-ready version of it: https://github.com/swirsz/fuzz-dojo

Scoder12:
> either share or synchronize the user database?
Custom dojos aren't a separate server; they're a group of modules on the same instance. 
> would it be possible to gate the dojo behind an IRB acceptance form that you must accept to continue?
not currently. this would need to be it's own PR

ConnorNelson:
What we would probably do for IRB is create a webpage somewhere on the sefcom site, and put the info to join a private dojo inside it (including a password) and state in that document that them joining  the private dojo constitutes consent.

PR #258, opened 2023-11-17
Make challenge composition granularly configurable

zardus:
This PR adds `INSTALL_*` variables for almost every piece of a container, allowing us to enable and disable them granularly. The inspiration was angr-management not shipping for and the kernel not compiling under ubuntu 23.04.


zardus:
Migration note: this changes `INSTALL_IDA=True` to `INSTALL_IDA=yes`, because these variables need to go into intermediate docker targets, and those can't have uppercase names. The problem is, if not updated manually in config.env, every previous installation of the dojo will fail to build the challenge container until `INSTALL_IDA` is updated. We can mitigate this somewhat by moving to `INSTALL_IDA_FREE` and deprecating `INSTALL_IDA`: then the failure mode would just be a missing IDA instead of a build failure. Thoughts?

zardus:
Okay, so the TLDR:

- the PR works
- the CI challenge is set to challenge-mini (because otherwise, due to the overlay bug, the CI runs out of disk space)
- non-mini challenge configurations would have failed on CI for the same reason before this change

I think we should merge this, because it's a big changeset to maintain. The layering bug affects us anyways, so there's no regression.

ConnorNelson:
Looks good to me, are we ready to merge?

PR #267, opened 2023-11-30
Discord: Add belting ceremony on solve

ConnorNelson:


ConnorNelson:
If/once we support multiple dojos getting the same belt, we should break out of the loop.

However, we also have to support the edge case where someone solves a challenge, and suddenly they have multiple belts ready to be awarded. We should award them in order.

And right now we just print instead of interface with Discord. I'd like to make sure this works first before we perform the discord operation (to avoid the terrifying case of 1000s of belts being suddenly awarded).

PR #273, opened 2023-12-29
[WIP] Improve performance of dojo/module ranking using Redis cacheing

adamdoupe:
Initial implementation. 

Remaining tasks:

- [x] Fetch all the ranking for the dojo/module, and use redis to store the results.
- [x] Invalidate cache when user account is updated
- [x] Invalidate cache when correct solve is made
- [x] Invalidate cache when updating/modifying a dojo
- [x] Create tests.

### DIFF
--- a/dojo_plugin/api/v1/dojo.py
+++ b/dojo_plugin/api/v1/dojo.py
@@ -22,7 +22,7 @@ from CTFd.utils.modes import get_model
 from CTFd.utils.security.sanitize import sanitize_html
 
 from ...models import Dojos, DojoMembers, DojoAdmins
-from ...utils.dojo import dojo_accessible, dojo_clone, load_dojo_dir
+from ...utils.dojo import dojo_accessible, dojo_clone, load_dojo_dir, dojo_route
 
 
 dojo_namespace = Namespace(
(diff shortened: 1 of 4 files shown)


ConnorNelson:
CTFd has utility functions `@cache.memoize(timeout)` and `cache.delete_memoized(func)` (`from CTFd.cache import cache`) which uses `redis`.

For invalidating, we can probably benefit from `@event.listens_for(SQLAlchemyModel, 'event_name')` (`from sqlalchemy import event`). I think we can hook all user updates / dojo updates / solve inserts, and invalidate there. That way we don't have to put code all over the place to do the invalidates.

adamdoupe:
So I dug into it a bit, the CTFd `cache` object is just this `flask-caching` library:
https://flask-caching.readthedocs.io/en/latest/

One problem with using `@cache.memoize` is that I _think_ that it includes the request path as part of the cache key: which means that the memoization doesn't persist across other requests. However, you can write a custom cache_key, so I'll look into that as well as dig into the source of the `flask-caching` library.

adamdoupe:
Ok, figured out the problem. Because we're using so much subclassing, you need to specify `propagate=True` on the sqlalchemy event. Otherwise the event won't fire when a subclass is changed (e.g., for Users there is the subclass of Admins)!

PR #275, opened 2024-01-08
Feat/belt changes

robwaz:
- Adds an icon field displayable on official dojos on the dojos page.  Use-case can include signifying which dojos correspond to which belt or uniquely identifying dojos in another manner.
@ConnorNelson - Is icon good as a field or should it be a data_field?  json in db feels dirty to me 🤷 

- Adds a green belt svg image.

- Updates index.html to link 2024 courses and declares earning belts is linked to TOPIC dojos.


dojo_plugin/utils/dojo/__init__.py
```diff
@@ -193,7 +195,7 @@ def assert_one(query, error_message):
 
     dojo_kwargs = {
         field: dojo_data.get(field, getattr(import_dojo, field, None))
-        for field in ["id", "name", "description", "password", "type", "award"]
+        for field in ["id", "name", "description", "password", "type", "award", "icon"]
```
ConnorNelson, line 198: This is meant to be covered by "award", which generically includes both belt awards and token awards. Displaying an icon should be determined based on that award.

dojo_plugin/models/__init__.py
```diff
@@ -58,6 +58,7 @@ class Dojos(db.Model):
     id = db.Column(db.String(32), index=True)
     name = db.Column(db.String(128))
     description = db.Column(db.Text)
+    icon = db.Column(db.String(256))
```
ConnorNelson, line 61: JSON is way more flexible to update/change. I know it's "dirty" but the flexibility makes it worth it. We make fields not-JSON if they're absolutely all over the code base/we need to be able to filter and sort based on it at the database level (for performance reasons).

ConnorNelson (changes requested):

dojo_plugin/utils/dojo/__init__.py
robwaz: What is the intended type of "award"?  If it's just "blue belt" the mapping logic for str<->file_path becomes hard-coded somewhere or couples the award name to the filename.  No?

PR #279, opened 2024-01-11
sshd/Dockerfile: moove docker installation from pip3 to apk

mudongliang:
We can install docker from apk instead of pip3


ConnorNelson:
Will this specifically install the python client and not a full on docker runtime? We only need a client.

mudongliang:
Oh, yeah, apk installs both client and daemon.

PR #285, opened 2024-01-16
git clone --depth 1

### LINKED ISSUE #284: Shallow clone when building
The challenge container clones a bunch of git repos (which itself is silly, b/c we are building all this stuff off master, but that’s an issue for another day). 

It does a full clone, which takes longer and uses more storage, when we only need the latest version to build.

Should be easily fixed with changing the clone to use depth 1.

Scoder12:
Closes #284


adamdoupe:
Awesome, thanks!

Issue #298, opened 2024-01-19
Feature: Dojo administration through UI


adamdoupe:
Request form community dojos: have ability to make a user an admin of a dojo.

PR #300, opened 2024-01-22
UI for promoting dojo members to dojo admins

zardus:


adamdoupe:
No tests?

adamdoupe:
Also closes #298.

zardus:
@ConnorNelson @adamdoupe alright, i need a fresh pair of eyes. I have no idea why the CSRF check seems to fail partway through this test (at https://github.com/pwncollege/dojo/blob/wip/promotions/test/test_running.py#L131)... Here's what happens:

image

AFAICT from googling, that's a CSRF check failure. It's definitely not CTFd's `@authed_only` or the `@dojo_route` or the only 403-returning check in https://github.com/pwncollege/dojo/blob/wip/promotions/dojo_plugin/api/v1/dojo.py#L89.

adamdoupe:
@zardus AH, I think I got this same error when doing flag submission!

I think the issue is sending data in `json` of the `post` argument, I think this does something with the content-type that passes this check in ctfd:
https://github.com/pwncollege/dojo/blob/wip/promotions/test/test_running.py#L271

ConnorNelson:
This is the CTFd logic where you are failing: https://github.com/CTFd/CTFd/blob/3536c4c423aab5b0b9c164af73f59fe29a4d7dcf/CTFd/utils/initialization/__init__.py#L322

ConnorNelson:
As @adamdoupe is saying, I think changing `data={}` to `json={}` will fix the issue. If you check out https://pwn.college/intro-to-cybersecurity/talking-web, you'll learn all about the difference between form data vs json data 😉

That being said, do we even need to test sending empty data? That check feels a little excessive to me. Should we check all of our endpoints for sending empty data?

zardus:
Thanks, guys! Freaking json...

I guess I have some coursework to review :P

zardus:
The empty test was to make sure it wasn't my own 403 triggering during my debugging. I removed it now

Issue #311, opened 2024-01-28
consider bringing back support for yaml-only dojos


zardus:
Having dojos in git repos is the only sane option for people making dojos, but it's a big PITA for testing. It would be nice if we can just POST some yaml to the `create` endpoint in our test scripts and have the dojos created. I'm thinking something like:

```
id: test-dojo
modules:
- id: test-module
  challenges:
  - id: test-challenge
    files:
    - run: "#!/opt/pwn.college/bash\ncat /flag"
```

I think this would really amp up our testing capabilities.

ConnorNelson:
It sounds like what you want is a utility function in our testcases for being able to more rapidly test something.

```py
def run_challenge_script(script: str, interpreter="bash") -> subprocess.Popen:
    ...

process = run_challenge_script("cat /flag")
assert process.stdout.read() == FLAG
```

I do not think it makes sense to support this as an actual dojo feature. I don't think we need several ways of being able to do the same thing.

zardus:
This is less for testing challenge functionality (though it would be useful for that), but more for testing dojo functionality, such as:

- challenge visibility
- awards (belts, emoji, etc)
- scoreboard
- dojo administration

It's all a pain to test right now, and this would help us avoid a bunch of test-x-dojo repos

ConnorNelson:
Ah I see. I still believe what you're asking for though is better testing infrastructure, and not a new user-facing interface that we now need to support.

For example, if we could initialize a dojo from local directories (no GitHub required).

Alternatively (additionally), we really need the ability to directly interface with CTFd land code. For example to query our db models in a sane way, or perform actions like starting a challenge without just always going through http endpoints.

zardus:
implemented in https://github.com/pwncollege/dojo/pull/315 (commit 70d5c5211ff869e09c3eef509fb5f47b8df60641)

PR #315, opened 2024-01-30
Changes to emoji eligibility and some testing upgrades

zardus:
This PR adds the following:

1. It moves example dojos into the Create page, removing them from the main listing.
2. It swaps the order of More Material and Courses on the main page. People looking for the courses will find the courses, but we should work hard to surface the community dojos.
3. It adds an admin-only endpoint for making dojos official.
4. It adds an admin-only endpoint for adding a dojo via yml (@ConnorNelson , i know you were concerned about this. Check out 70d5c5211ff869e09c3eef509fb5f47b8df60641; it's pretty small)
5. It makes example dojos no longer award emoji


zardus:
Closes https://github.com/pwncollege/dojo/issues/311

PR #323, opened 2024-02-04
challenge/bash.bashrc: fix an issue "less command not found"

mudongliang:
Now challenge/bash.bashrc uses less command to show the /challenge/README.md. However, this command is not installed unless you use challenge-full.

Fix this by checking less command is installed or not.


zardus:
Thanks!

PR #324, opened 2024-02-05
sshd/active.py: add escape sequence when user enters a new chall

bravosixgoingdark:
The program would clear the terminal every time a new chall is activated. Tested on both Linux and Windows (cmd and powershell).


zardus:
Thanks!

bravosixgoingdark:
> Thanks!

No problems, my first time ever contribution on github lol

PR #325, opened 2024-02-06
Belts: Fix ceremony

ConnorNelson:


ConnorNelson:
Previously, `get_belt_awards` would (incorrectly) attempt to get belts in db, and computed belts, simultaneously. This functionality was only used in the service of `update_awards`. This PR makes the logic simpler (look at that net -17 LOC!) by just doing the award updating logic entirely inside `update_awards`. This function now also handles the Discord logic, making our CTFd solve hook simpler.

zardus:
This looks reasonable, but won't it make a discord API query to check for every belt role every single time any user submits any flag? I guess discord is probably fine with that?

zardus:
Actually, a few comments incoming.

dojo_plugin/utils/awards.py
```diff
@@ -74,13 +60,26 @@ def get_belts():
...
-    current_belts = get_user_belts(user)
-    for belt in current_belts:
-        belt_award = Belts.query.filter_by(user=user, name=belt).first()
-        if belt_award:
+    current_belts = [belt.name for belt in Belts.query.filter_by(user=user)]
+    discord_user = get_discord_user(user.id)
+    discord_roles = get_discord_roles()
+    for belt, dojo_id in BELT_REQUIREMENTS.items():
+        belt_role = belt.title() + " Belt"
+        missing_role = discord_user and discord_roles.get(belt_role) not in discord_user["roles"]
```
zardus, line 68: As mentioned, I think this fires for every belt role every time any user solves anything...

dojo_plugin/utils/awards.py
```diff
@@ -74,13 +60,26 @@ def get_belts():
...
+        if belt in current_belts and not missing_role:
             continue
-        db.session.add(Belts(user=user, name=belt))
-        db.session.commit()
+        dojo = Dojos.query.filter(Dojos.official, Dojos.id == dojo_id).first()
+        if not (dojo and dojo.completed(user)):
```
zardus, line 72: This would redundantly fire even if a user already has the belt role, eating some small amounts of resources.

dojo_plugin/utils/awards.py
```diff
@@ -74,13 +60,26 @@ def get_belts():
...
+            continue
```
zardus, line 73: This seems to make it impossible for hackers that have a legacy belt award (e.g., from a DB sync or for some other reason) that no longer satisfies modern requirements and haven't received the discord role to ever get the discord role.

dojo_plugin/utils/awards.py
```diff
@@ -74,13 +60,26 @@ def get_belts():
...
+        if belt not in current_belts:
+            db.session.add(Belts(user=user, name=belt))
+            db.session.commit()
+        if missing_role:
+            user_mention = f"<@{discord_user['user']['id']}>"
+            message = f"{user_mention} earned their {belt_role}! :tada:"
+            print(message, flush=True)
```
zardus, line 80: do we need to print this out? Is that for local logging?

zardus (commented):
I like everything living in update_awards, but I think the logic should be:

1. Figure out what belts the user should have
2. Figure out what Belt awards need to be created
3. Figure out what Discord roles need to be granted

In that order, probably? The interleaving that's there right now makes this hard to get right without wasting calls.

We also need to keep in mind legacy requirements. People need to be able to get future belts even if their past belts were earned under earlier requirements.

Finally, I just noticed that this looks like it gets rid of the cumulative belt behavior altogether (e.g., with this PR, people can earn a blue belt directly).

dojo_plugin/utils/awards.py
ConnorNelson: Sure does, just like before.

dojo_plugin/utils/awards.py
ConnorNelson: Yeah, you're right; the existence of the belt award should definitely be saving us from doing this work.

dojo_plugin/utils/awards.py
ConnorNelson: True

dojo_plugin/utils/awards.py
ConnorNelson: Nope!

ConnorNelson (commented):
The next PR should address everything we can address.

dojo_plugin/utils/awards.py
```diff
@@ -74,13 +60,30 @@ def get_belts():
...
-    for belt in current_belts:
-        belt_award = Belts.query.filter_by(user=user, name=belt).first()
-        if belt_award:
+    current_belts = [belt.name for belt in Belts.query.filter_by(user=user)]
+    for belt, dojo_id in BELT_REQUIREMENTS.items():
+        if belt in current_belts:
+            continue
+        dojo = Dojos.query.filter(Dojos.official, Dojos.id == dojo_id).first()
+        if not (dojo and dojo.completed(user)):
             continue
```
zardus, line 69: I think the remaining issue is that, due to this continue, belts are no longer cumulative. Shouldn't this be a `break`?

zardus (changes requested):

dojo_plugin/utils/awards.py
ConnorNelson: Good call

zardus:
We now have at least some testcases for belts. Let's see if they pass on this branch!

PR #333, opened 2024-03-01
Add some apt packages that 543 web challenges require.

ltfish:


zardus:
Sanity check: these get explicitly started by the init script, right?

PR #336, opened 2024-03-03
set a default host for the nginx proxy

zardus:
A common issue people run into is trying to access the dojo by IP when setting up a self-hosted instance. This PR fixes that by adding a default proxy route.


gogo2464:
does it requires: `export DOJO_HOST=127.0.0.1`

PR #337, opened 2024-03-04
add binja free

robwaz:
Adds Binary Ninja free to challenge image along with icon to xfce panel and application launcher if `INSTALL_BINJA_FREE` param is `yes`.


zardus:
We should explicitly disable binja by default, similar to what we do for IDA (here: https://github.com/pwncollege/dojo/blob/master/script/container-setup.sh#L36). The idea is that people should have to explicitly enable anything commercial like that.

Now that I think about it, we should also do that for Windows when that becomes `INSTALL_WINDOWS`-toggleable

robwaz:
I'll default `INSTALL_BINJA_FREE` to `no` right now.  I'll update similar with windows in the 598-windows branch as some of that is getting reworked anyways

PR #339, opened 2024-03-05
challenge/Dockerfile: add a package dwarves

mudongliang:
The linux kernel 5.4 compilation encounters the following warning:

.tmp_vmlinux.btf: pahole (pahole) is not available

Fix it by adding a package dwarves


mudongliang:
BTW, this warn can be also fixed by disabling CONFIG_DEBUG_INFO_BTF. But I think this configuration should be enabled by default.

PR #341, opened 2024-03-05
Added verification for port 465 in the mail_tls parameter.

wumingzhilian:
In the config.py file, validation for port 465 has been added, as port 465 is for SSL/TLS.


ConnorNelson:
Thanks!

PR #342, opened 2024-03-07
Add selenium-driven testcases that test user-facing functionality

zardus:
The recent practice mode issue with the welcome dojo got me thinking that the welcome dojo actually makes for a really nice smoke test of pwn.college's user-facing functionality. This PR adds initial capabilities to test this using selenium.


zardus:
Dedicating this one to @adamdoupe

PR #347, opened 2024-03-07
Workspace: Enter ssh through ssh-entrypoint

ConnorNelson:
VSCode and Desktop can be easily tracked, SSH is actually harder. This fixes that.

### DIFF
--- a/challenge/Dockerfile
+++ b/challenge/Dockerfile
@@ -587,6 +587,7 @@ COPY --link --from=essentials / /
 COPY --link --from=builder-pwn.college /opt/pwn.college /opt/pwn.college
 
 RUN <<EOF
+    ln -sf /bin/bash /opt/pwn.college/ssh-entrypoint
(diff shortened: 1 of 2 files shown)


zardus:
Where is ssh-entrypoint itself?

ConnorNelson:
It's just a symlink to bash, the objective is to make someone using ssh can be identified by `/proc/PID/cmdline`.

VSCode can be identified by the existence of a `/usr/lib/code-server/lib/node --dns-result-order=ipv4first /usr/lib/code-server/lib/vscode/out/bootstrap-fork --type=extensionHost --transformURIs --useHostProxy=false`.

Desktop can be identified by >1 existence of `/usr/bin/python3 /usr/bin/websockify --web /usr/share/novnc/ dojo-user:6081 --unix-target=/tmp/.dojo/vnc/socket`.

zardus:
Color me embarrassed...

zardus (approved):

PR #352, opened 2024-03-09
Allow selectively disabling practice mode

zardus:
Practice mode makes it difficult to have challenges with static secret sauce. @ltfish and others have workarounds where they detect practice mode and delete secrets, but we should make this easier.

A side benefit of allowing disabling of practice mode is that we'll be able to disable practice mode on the early levels of the welcome dojo, until practice mode is discussed.

Deployment caveat: this adds a column to DojoChallenges.

### DIFF
--- a/dojo_plugin/utils/dojo.py
+++ b/dojo_plugin/utils/dojo.py
@@ -241,12 +241,14 @@ def dojo_from_spec(data, *, dojo_dir=None, dojo=None):
             return cls(start=start, stop=stop)
 
     _missing = object()
-    def shadow(attr, *datas, default=_missing):
+    def shadow(attr, *datas, default=_missing, default_dict=None):
         for data in reversed(datas):
             if attr in data:
                 return data[attr]
         if default is not _missing:
             return default
+        elif default_dict and attr in default_dict:
+            return default_dict[attr]
         raise KeyError(f"Missing `{attr}` in `{datas}`")
 
     def import_ids(attrs, *datas):
@@ -260,8 +262,8 @@ def dojo_from_spec(data, *, dojo_dir=None, dojo=None):
                 DojoChallenges(
                     **{kwarg: challenge_data.get(kwarg) for kwarg in ["id", "name", "description"]},
                     image=shadow("image", dojo_data, module_data, challenge_data, default=None),
-                    allow_privileged=shadow("allow_privileged", dojo_data, module_data, challenge_data, default=True),
-                    importable=shadow("importable", dojo_data, module_data, challenge_data, default=True),
+                    allow_privileged=shadow("allow_privileged", dojo_data, module_data, challenge_data, default_dict=DojoChallenges.data_defaults),
+                    importable=shadow("importable", dojo_data, module_data, challenge_data, default_dict=DojoChallenges.data_defaults),
                     challenge=challenge(module_data.get("id"), challenge_data.get("id")) if "import" not in challenge_data else None,
                     visibility=visibility(DojoChallengeVisibilities, dojo_data, module_data, challenge_data),
                     default=(assert_import_one(DojoChallenges.from_id(*import_ids(["dojo", "module", "challenge"], dojo_data, module_data, challenge_data)),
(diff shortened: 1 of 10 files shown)


zardus:
One note: currently, nothing prevents people from importing the challenge and re-enabling practice mode on the import. We should probably disallow overriding that on import.

zardus:
Okay, now we have an `unimportable` attribute, and both `unimportable` and `practice_enabled` have been moved to json fields in the json data, so no more extra DB columns, making deployment easy. To facilitate existing entries without these keys, I added a `data_defaults` dict to dojos, modules, and challenges that has defaults (currently just for importable and practice_enabled).

I think this should be ready to go.

ConnorNelson:
Can we change `practice_enabled` to `allow_privileged`? We are transitioning away from "practice", towards "privileged" (it's even the term used in the DOJO paper).

---

I will also push slightly against this overall direction. I think one of the important ideas in pwn.college is highly introspect-able challenges as a matter of challenge design. In my opinion, black box components of a challenge design is almost-always a bad education-first decision. Obviously there's a place for it (static secret data, *learning* about black-box related thinking) , and obviously `.init` already allows this to happen, but I think we should largely caution against black box components in a challenge. I think it is unfortunately often a crutch for poor challenge design focused on adding difficulty at a disproportionate cost to education. There should be an *extreme* emphasis on students being able to see for themselves how a system works. 

For example, if I have some PHP web server, I should be able to easily edit the challenge (in privileged mode) to add additional logging, and see that quickly reflected. Now I imagine some challenge where you leak out the source code as the first step, but you don't have quick introspection capabilities, because there's no mechanism to (for example) modify that source. Instead students have to redeploy the web server themselves, get the configuration details right, etc. Are they learning something? Sure. Are those frustrating barriers worth it though? I guess it depends on what the challenge is trying to teach. Unfortunately, this feature further encourages that style of design as a first-class idea in pwn.college. I worry about losing the *extreme* emphasis on introspect-ability as a matter of education-first challenge design.

I'm not a hard no on this direction, just something I wanted to voice.

zardus:
Man, it's like you have a PhD in cybersecurity education or something...

FWIW, I agree that the recommendation should be strongly in the direction of availability of privileged mode. But there are two legitimate, educationally-sound usecases for this:

1. I'm not sure there's another reasonable way to teach, for example, black-box web scanning. The point there *is* the black box.

2. Keeping in mind that pwn.college's educational unit is a Module, not necessarily an individual Challenge (e.g., series of challenges build up to convey a concept), I see something like this being very useful for simplifying, for example, the blind-reversing yan85 challenge. We had to implement a whole runtime randomization thing for yan85 there, and it's rather hacky, but with this functionality, we could instead have that module as two levels: the first is a statically-randomized yan85 that you can access (this already exists, in fact), and the second is just another instance, randomized differently, that you _cannot_ access. If you need privileged access, you can do it in the first one --- they're the same challenge, the difference is just that you don't have access to the binary. In this example, I think the ability to disabled privileged access would actually end up helping the learning ramp, since students will have just one concept (inferring binary functionality based on output) to tackle instead of two (inferring binary functionality based on output _and_ understanding the runtime randomization code.

That being said, I agree that we should strongly push in the allow_privileged direction. If you feel strongly enough about it, we could make this functionality only available for admin-owned dojos? IMO, it's okay as is (with a philosophy guide), but I'm not super against gating it.

dojo_plugin/utils/dojo.py
```diff
@@ -63,6 +65,8 @@
         **VISIBILITY,
 
         Optional("image"): IMAGE_REGEX,
+        Optional("allow_privileged"): bool,
+        Optional("unimportable"): bool,
```
ConnorNelson, line 69: Let's change `unimportable` to `importable` (with a default of `True`, of course). I know that using this field explicitly means you're going to be disabling importability, but spending any time thinking about a double negative is just always annoying. I think `importable: False` is much more readable than `unimportable: True`.

One other naming thought would be `public_importable` if we think there's any realistic chance in the future we need dojo admins to be able to import from one of their other dojos which they are an admin in, but no one else should be able to. For example, if we actually end up using this in a belted dojo, and then want to use it in a course dojo. Some bypass (probably this outlined common admin idea) would be necessary.

zardus:
I opened #359 and #360 to track the docs part of this. Making the `importable` change (I like it better than `public_importable`; we can cross that bridge when we come to it), and will merge when I get tests to pass!

PR #356, opened 2024-03-10
Add a testcase for the desktop

zardus:


zardus:
Downside of this: longer test times. We can re-shorten it by switching to xterm (and not needing to build all the xfce stuff), with the downside of that being that the `Using the GUI Desktop` challenge's check doesn't catch that. Probably worth it, though...

PR #357, opened 2024-03-11
Make the default scoreboard monthly

zardus:
We'd discussed the annoyance of the static scoreboards. This PR makes the scoreboards monthly (by default), moves the duration controls into the scoreboard macro itself, and bolds the currently-active one.


zardus:
yolo

PR #366, opened 2024-03-21
Windows rebuild

robwaz:
Addresses a few issues with the current Windows implementation and hopefully makes future dev less time-intensive.

Major Changes:
- Removed asynchronous Rust challenge-proxy for synchronous challenge-proxy written in C to address challenge IO bugs.
- Added windows build steps that allow for using cached windows installation.
- Added `post_install.sh` and `post_install.ps1` which leverage cached windows installation and should be primary location for future changes to Windows VM.


robwaz:
I deny the red checkmark of "operation cancelled"!

PR #369, opened 2024-03-27
Dojo: Increase admin visibility

ConnorNelson:
Motivated from the Discord:

> Hi Dr. Kanak, I have a question about the owner/creator of one dojo and the dojo admin. It seems some functionability (e.g., make the dojo official) needs both roles satisfied.
> And I encountered a wired issue in making one dojo official: The owner A creates a dojo. He/She can see this dojo in its management webpage and have “Promote to Admin”, but does not have "Make the dojo official". However, the dojo admin cannot access this dojo with admin privilege.
> So we need to resort to database modification about this operation
>


mudongliang:
@ConnorNelson it works as expected. For dojo admin, you can go to "Admin Panel" -> "Dojos" -> "Dojo you would like to operate on" -> "Dojo Admin Panel". Then make it official

PR #372, opened 2024-03-27
Add dojo-LFS support

zardus:
github doesn't allow very large files to be checked into git repos, which is problematic for some dojos that need, e.g., different kernel builds. This adds support for downloading files post-clone.

### DIFF
--- a/dojo_plugin/api/v1/dojo.py
+++ b/dojo_plugin/api/v1/dojo.py
@@ -25,45 +25,30 @@ from CTFd.utils.modes import get_model
 from CTFd.utils.security.sanitize import sanitize_html
 
 from ...models import Dojos, DojoMembers, DojoAdmins, DojoUsers, Emojis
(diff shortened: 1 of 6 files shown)


zardus:
@ConnorNelson based on your feedback, I disabled the large file updating. Dropbox doesn't have any way to tell a file hash without downloading it, so I figured we err on the side of just not redownloading tons of files (and timing out the web request). This also allows us, for now while we have this hacky solution of doing the clone in the web request context, to pull one file at a time.

Would you feel better about all these shenanigans if I made this feature admin-only?

Eventual TODOs:
- some sort of LFS file management (delete, see hash, etc) via the admin page

ConnorNelson:
Yeah I appreciate this being admin-only until we flesh this out and see some more interest in this feature.

If Dropbox does not expose the hash given a url, we could also track url/path mappings in a `.lfs` file or something, to prevent duplicating work.

zardus:
I figure we just don't re-download files that are already there, and that'll be sufficient for that. There's no real way to deal with the hashing thing using any reasonable dropbox APIs, so if we ever want to update LFS files, we'll need to expose that via the web UI some day

PR #374, opened 2024-03-28
limit windows start cmd

f4-u57:
limit `windows start` cmd to Windows challenges by checking for any file that ends in `.exe` in `/challenge/` when `windows start` is used.

### DIFF
--- a/challenge/windows/windows
+++ b/challenge/windows/windows
@@ -61,6 +61,10 @@ def is_privileged():
     return os.getuid() == 0 or "sudo" in groups
 
 
+def has_windows_challenges():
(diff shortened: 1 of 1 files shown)


ConnorNelson:
Is there a specific reason we want to limit the windows vm from starting without a `/challenge/*.exe`? For example, sometimes it's nice to start a linux vm in a random challenge while messing around with things. Do you have specific concerns here?

f4-u57:
Window VM shouldn't be trusted because Windows is insane

f4-u57:
#384 guarantees the Windows VM is only accessible after the configuring stages so this PR can be closed.

PR #376, opened 2024-03-28
configure gdb plugins in the challenge initialization

mudongliang:


ConnorNelson:
I don't think we want to force that specific `.gdbinit` as a default for users. 

I would be fine if we wanted to somehow better convey that `/opt/pwndbg` and `/opt/gef` exist and can be used with an `echo 'source /opt/pwndbg/gdbinit.py' > ~/.gdbinit` or `echo 'source /opt/gef/gef.py' > ~/.gdbinit`. I'm not totally sure what that better conveyance of information would look like; probably it could make sense as a challenge in the `debugging-refresher` module.

mudongliang:
Yeah, this gdbinit looks really complicated for newbies.

mudongliang:
I would like to enable pwndbg and find that .gdbinit includes "source /opt/pwndbg/gdbinit.py". So I post this PR

mudongliang:
And there is another issue in gef. It has changed the installation method.

ConnorNelson:
I just tested both of the above `echo` commands for using pwndbg/gef, and they both seem to work.

mudongliang:
I have an idea. Maybe we can craft a challenge - pwndbg, gef to tell newbies that, we have installed many fancy plugins, you can enable anyone you like.

mudongliang:
like many challenges in the welcome-dojo, this is an introduction for the functionability of dojo platform

ConnorNelson:
This is a gdb capability. It should be discussed in the gdb module.

mudongliang:
This is just an example. I checked dojos in pwncollege, Debugging-fresher is a good choice.

mudongliang:
Let's close this PR

PR #377, opened 2024-03-29
Change dojo update to do a rebase pull and be more careful about it

zardus:
Inspired by our unstaged commits on the central instance, this tries to make `dojo update` less risky... Requires all changes to be locally committed, though.

Not married to this approach; but it's probably better than what we have now

### DIFF
--- a/script/dojo
+++ b/script/dojo
@@ -21,7 +21,18 @@ DOJO_DIR=/opt/pwn.college
 case "$ACTION" in
     # HELP: update: update dojo files (warning: does `git pull`), rebuild containers, and restart any changed services
     "update")
-        git pull
+        git config --global --add safe.directory "$DOJO_DIR"
+        if ! git diff --exit-code
+        then
+            echo "ABORTING: working directory is not clean"
+            exit 1
+        fi
+        if ! git pull -r
+        then
+            echo "ABORTING: 'git pull -r' failed"
+            git rebase --abort
+            exit 1
+        fi
         dojo sync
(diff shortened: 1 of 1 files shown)


script/dojo
```diff
@@ -21,7 +21,18 @@ DOJO_DIR=/opt/pwn.college
 case "$ACTION" in
     # HELP: update: update dojo files (warning: does `git pull`), rebuild containers, and restart any changed services
     "update")
-        git pull
+        git config --global --add safe.directory "$DOJO_DIR"
```
ConnorNelson, line 24: Is this required?

adamdoupe:
I don't like this approach. We should move to have everything in non-versioned files like `config.env` that was the rest of the system should be identical for what we use in prod vs dev.

ConnorNelson:
Just to confirm, this doesn't require us to commit untracked files/directories, correct? I would be against that if this does.

Otherwise, I'm pretty neutral. It does seem like a nice idea to ensure someone doesn't do an update that possibly leads to some merge conflict. On the other hand, if on production I, for example, disable the scoreboard, but then want to bring in some commits, this is going to make that more annoying.

I wonder if there's a way to check if the `git pull -r` would succeed before actually doing it, and let the pull through if it would, and otherwise yell.

zardus:
I also like not having local changes, of course! If we can get to that point, then this is completely unnecessary :-)

zardus:
We got to that point!

PR #378, opened 2024-03-29
Linux VM - Support challenge defined bzImage and vmlinux

robwaz:
@ConnorNelson Quick sanity check?


ConnorNelson:
Let's do no glob. Otherwise looks good to me.

PR #379, opened 2024-03-29
Allow better support of externally hosted DBs

adamdoupe:
This way, the docker-compose.yml file does not need to change to support an external db.

Now the db container will detect the DB_EXTERNAL environment variable, and if it is not `no` then the db container will just sleep rather than start up mysql.

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,6 +1,5 @@
 FROM mariadb:10.4.12
 
 COPY ./start.sh /start.sh
-RUN chmod +x /start.sh
 USER mysql
 
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,0 +1,6 @@
+FROM mariadb:10.4.12
+
+COPY ./start.sh /start.sh
+RUN chmod +x /start.sh
+USER mysql
+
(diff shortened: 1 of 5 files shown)


script/container-setup.sh
```diff
@@ -39,6 +39,7 @@ define DB_HOST db
 define DB_NAME ctfd
 define DB_USER ctfd
 define DB_PASS ctfd
+define DB_EXTERNAL no # change to anything but no and the db container will not start mysql
```
ConnorNelson, line 42: My only real thought here is that `DB_EXTERNAL` is redundant with `DB_HOST` being vs not being `db`. As a result, `DB_HOST=db; DB_EXTERNAL=yes` would lead to a weird state.

I would personally lead towards less config options, but up to you if you prefer this explicit config option name.

db/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mariadb:10.4.12
+
+COPY ./start.sh /start.sh
+RUN chmod +x /start.sh
```
ConnorNelson, line 4: Wait is this actually required? Does docker not COPY with the existing executable permission?

script/container-setup.sh
adamdoupe: I didn't like the idea of further hardcoding in the infra of what the container name should be. 

In general I prefer explicit configuration vs. magic configuration. But it is redundant and there is a possibility of these weird states.

PR #384, opened 2024-03-31
Windows VM improvments

f4-u57:
Summary of changes:

1. The VNC service should not start with Windows because the VM should not be accessible by the user during the configuring steps in the `startup.ps1`. It should start at the end of startup.ps1 as intended. The `tvnserver` service is starting with Windows because, by default, its service type is `Automatic` which means it will start with Windows. The fix is to set the service type to Manual which means the service will only start if involved to start. I also added setting the service type of the `sshd` service to Manual even though it's already Manual by default just for clarity. So `Set-Service -Name sshd -StartupType Manual` is just there for clarity, it's not needed. After this change, the user will be able to connect to the VM via VNC after the configuring steps in `startup.ps1` instead of when Windows starts.

2. The `Stop-Computer` command will not be able to shut down Windows if there are users logged in and `-force` fixes that. 

3. Logoff by session name, `console`, instead of session ID since session ID can change when logon or logoff occurs.

I verified these changes locally, the vnc change from the user's perspective, and parsed the Windows Event log.

Windows Events Log parse after this change. It shows vnc server starts at end of startup.ps1 (at the `Starting ssh and vnc servers` line)
```
---------------------------------------------------------------------------------------------------------------------------
2024-03-31 04:56:11.801355 UTC - System start                                                                                                                                                           
---------------------------------------------------------------------------------------------------------------------------                                                                                                                                                                                                                                                     
2024-03-31 04:56:32.025317 UTC - OpenSSH - Server listening on :: port 22.                                                                                                                              
2024-03-31 04:56:32.025441 UTC - OpenSSH - Server listening on 0.0.0.0 port 22.                                                                                                                         
---------------------------------------------------------------------------------------------------------------------------                                                                                                                                                                                                                                               
2024-03-31 04:56:28.990228 UTC - startup-script - Started                                                                                                                                               
2024-03-31 04:56:29.974674 UTC - startup-script - Mounted X,Y,Z drive                                                                                                                                   
2024-03-31 04:56:31.818441 UTC - startup-script - Starting ssh and vnc servers                                                                                                                          
2024-03-31 04:56:32.193310 UTC - tvnserver - Service has been started successfully                                                                                                                      
2024-03-31 04:56:32.443436 UTC - startup-script - Finished                                                                                                                                              
2024-03-31 04:56:39.693353 UTC - tvnserver - Authentication passed by 10.0.2.2
---------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/challenge/windows/post_install.ps1
+++ b/challenge/windows/post_install.ps1
@@ -25,6 +25,4 @@ cl challenge-proxy.c
 
 
 # -- shutdown --
-Set-Service -Name sshd -StartupType Manual
-Set-Service -Name tvnserver -StartupType Manual
 Stop-Computer -computername localhost
--- a/challenge/windows/post_install.ps1
+++ b/challenge/windows/post_install.ps1
@@ -25,4 +25,4 @@ cl challenge-proxy.c
 
 
 # -- shutdown --
-Stop-Computer -computername localhost
+Stop-Computer -computername localhost -force
--- a/challenge/windows/setup.ps1
+++ b/challenge/windows/setup.ps1
@@ -253,5 +253,9 @@ Add-Content -Path $env:windir\System32\drivers\etc\hosts -Value "`n$ip`tpublic-l
 Copy-Item A:\config_startup.ps1 -Destination "C:\Program Files\Common Files\startup.ps1"
 & schtasks /create /tn "dojoinit" /sc onstart /delay 0000:00 /rl highest /ru system /tr "powershell.exe -file 'C:\Program Files\Common Files\startup.ps1'" /f
 
+# config services' StartupType to start when Start-Service is called or manually started (Manual) instead of start with Windows (Automatic)
+Set-Service -Name sshd -StartupType Manual
+Set-Service -Name tvnserver -StartupType Manual
+
 # -- shutdown --
 Stop-Computer -computername localhost -force
--- a/challenge/windows/setup.ps1
+++ b/challenge/windows/setup.ps1
@@ -109,7 +109,7 @@ function EnableWmiRemoting($namespace) {
         throw "GetSecurityDescriptor failed: $($output.ReturnValue)"
     }
     $acl = $output.Descriptor
-    
+
     $computerName = (Get-WmiObject Win32_ComputerSystem).Name
     $acc = Get-WmiObject -Class Win32_Group -Filter "Domain='$computerName' and Name='Users'"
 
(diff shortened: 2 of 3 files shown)


challenge/windows/post_install.ps1
```diff
@@ -25,6 +25,4 @@ cl challenge-proxy.c
 
 
 # -- shutdown --
-Set-Service -Name sshd -StartupType Manual
-Set-Service -Name tvnserver -StartupType Manual
```
ConnorNelson, line 29: "The fix is to set the service type to Manual which means the service will only start if involved to start."

To make sure I understand these changes, and to document them here, can you please clarify this. It looks like in the past it was already "Manual". Instead you are moving this from `post_install.ps1` to `setup.ps1`. Do we have a good understanding of when `tvnserver` (and `sshd`) need to start in order to guarantee the hacker user doesn't have admin privileges? What was going wrong before, and why does this fix it? Again, I'd like to document this since I have a very weak mental model about what is going on here.

f4-u57: My bad, typo here: "The fix is to set the service type to Manual which means the service will only start if ~involved~ invoked to start."

The two `Set-Service` lines in `post_install.ps1` changed the two services' type to Manual but it does not persist after reboot, so `tvnserver` will still start Windows boot since its service type is still Automatic.  I haven't looked into why the changes did not persist in `post_install.ps1` yet, so I just moved them to `setup.ps1` which persists after reboot. Also, the `sshd` service's type by default is already Manual,  so the `Service -Name sshd -StartupType Manual` line is not needed, it's just there to be verbose about it needs to be Manual.

We want the `tvnserver` and `sshd` to start after the logout in `startup.ps1`, it guarantees the hacker user doesn't have admin privileges because if the hacker user is currently logged in and group change is made on the hacker user, a logout is needed to reflect that change.  

https://github.com/pwncollege/dojo/blob/661322679a27b3d43cbf2596395615d6339ed8b6/challenge/windows/startup.ps1#L39-L42

What's wrong before is `Start-Service tvnserver`  is not actually controlling when to start the `tvnserver` service since its type is Automatic and the `tvnserver` service will just start with Windows boot. The fix is to change it to Manual and make sure it persists after reboot, then the `tvnserver` service will start when `Start-Service tvnserver` is called. 

Issue before and order of events:
1. Windows boot up
2. `tvnserver` service start (right after boot up and the user can connect right here before )
3. `startup-script.ps1` runs
4. `sshd` service start (at the end of `startup-script.ps1`)

Current (this PR) and order of events:
1. Windows boot up
2. `startup-script.ps1` runs
3. `tvnserver` service start (at the end of `startup-script.ps1`)
4. `sshd` service start (at the end of `startup-script.ps1`)

Scoder12: I should have mentioned this earlier but the changes don't persist because because for some reason the actual service creation happens asynchronously, so where the `Set-Service` used to be it would happen too early and just silently error. You can see the error message if you're watching the logs as it builds

PR #387, opened 2024-04-02
add invite dojo feature

wumingzhilian:
add invite dojo feature


ConnorNelson:
This already exists as the `Share` button?

wumingzhilian:
> This already exists as the `Share` button?

Oh! Sorry, it's my fault. Just saw the share function.

PR #392, opened 2024-04-06
fix display of emoji not attached to a dojo

zardus:
the recent performance fixes broke non-dojo-specific emojis from displaying. This fixes it.


zardus:
Ah, looks like we're hitting time limits even without the XFCE build change

PR #398, opened 2024-04-14
Add functionality for multiple ssh keys

### LINKED ISSUE #396: dojo: Moar SSH Keys
As mentioned on the discord, folks often use multiple SSH keys, so it’d be great if the dojo allowed adding multiple SSH keys through the web interface.

The usecase is that folks don’t want to transfer the private key between devices (home/work) and would like to specify multiple keys.

XploitBengineer:
This is a proposal to fix #396. This functionality would be really useful for me so when I saw the issue be created I thought I'd make a start on it. Still WIP as I'm not sure if it could cause a breaking change.

The main difference is a change to the structure of the SSH Keys table. To allow users to have multiple keys, I don't think the user id can be the public key anymore. Since currently public keys must be unique currently, this could become the new primary key, but for now I just made a row id field. Making this change also breaks the current database, but I'm not sure how else to implement the feature.

I've also done the following

- Added a new API endpoint, for deleting an API key.
- Added a form in the settings page for displaying and removing keys, it looks absolutely horrible at the moment though, need to make it prettier.

It looks like `auth.py` is already capable of adding multiple keys to the challenge container, so I didn't make any changes to this.

### DIFF
--- a/dojo_plugin/api/v1/ssh_key.py
+++ b/dojo_plugin/api/v1/ssh_key.py
@@ -18,11 +18,9 @@ ssh_key_namespace = Namespace(
 @ssh_key_namespace.route("")
 class UpdateKey(Resource):
     @authed_only
-    def patch(self):
-
+    def post(self):
         data = request.get_json()
         key_value = data.get("ssh_key", "")
-
 
         if key_value:
             key_re = "ssh-(rsa|ed25519|dss) AAAA[0-9A-Za-z+/]+[=]{0,2}"
@@ -43,7 +41,6 @@ class UpdateKey(Resource):
             key = SSHKeys(user_id=user.id, value=key_value)
             db.session.add(key)
             db.session.commit()
-
         except IntegrityError:
             db.session.rollback()
             return (
@@ -53,11 +50,8 @@ class UpdateKey(Resource):
 
         return {"success": True}
     
-    
-@ssh_key_namespace.route("/delete")
-class DeleteKey(Resource):
     @authed_only
-    def patch(self):
+    def delete(self):
         data = request.get_json()
         key_value = data.get("ssh_key", "")
 
--- a/dojo_plugin/api/v1/ssh_key.py
+++ b/dojo_plugin/api/v1/ssh_key.py
@@ -55,7 +55,7 @@ class UpdateKey(Resource):
     
     
 @ssh_key_namespace.route("/delete")
-class UpdateKey(Resource):
+class DeleteKey(Resource):
     @authed_only
     def patch(self):
         data = request.get_json()
--- a/dojo_plugin/api/v1/ssh_key.py
+++ b/dojo_plugin/api/v1/ssh_key.py
@@ -19,8 +19,10 @@ ssh_key_namespace = Namespace(
 class UpdateKey(Resource):
     @authed_only
     def patch(self):
+
         data = request.get_json()
         key_value = data.get("ssh_key", "")
+
 
         if key_value:
             key_re = "ssh-(rsa|ed25519|dss) AAAA[0-9A-Za-z+/]+[=]{0,2}"
@@ -41,6 +43,7 @@ class UpdateKey(Resource):
             key = SSHKeys(user_id=user.id, value=key_value)
             db.session.add(key)
             db.session.commit()
+
         except IntegrityError:
             db.session.rollback()
             return (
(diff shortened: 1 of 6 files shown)


dojo_plugin/pages/settings.py
```diff
@@ -38,7 +38,7 @@ def settings_override():
         "settings.html",
         user=user,
         tokens=tokens,
-        ssh_key=ssh_key,
+        ssh_key=[i.value for i in ssh_keys],
```
ConnorNelson, line 41: Rename this to `ssh_keys`

dojo_plugin/pages/settings.py
```diff
@@ -17,8 +17,8 @@ def settings_override():
     user = get_current_user()
     tokens = UserTokens.query.filter_by(user_id=user.id).all()
 
-    ssh_key = SSHKeys.query.filter_by(user_id=user.id).first()
-    ssh_key = ssh_key.value if ssh_key else None
+    ssh_keys = SSHKeys.query.filter_by(user_id=user.id).all()
```
ConnorNelson, line 20: `SSHKeys.query.filter_by(user_id=user.id).with_entities(SSHKeys.value).all()`
Minor, but figured I'd mention it; this gets you the data you want.

dojo_plugin/api/v1/ssh_key.py
```diff
@@ -19,9 +19,11 @@
 class UpdateKey(Resource):
     @authed_only
     def patch(self):
```
ConnorNelson, line 21: Let's actually change `patch` to `post`.

dojo_plugin/api/v1/ssh_key.py
```diff
@@ -53,3 +52,45 @@ def patch(self):
             )
 
         return {"success": True}
+    
+    
+@ssh_key_namespace.route("/delete")
+class DeleteKey(Resource):
```
ConnorNelson, line 58: Let's not create another Resource, put this as a method named `delete` under the original resource. This gives us our CRUD operations for POST/DELETE to add/remove.

dojo_plugin/models/__init__.py
```diff
@@ -679,8 +679,9 @@ class DojoResourceVisibilities(db.Model):
 
 class SSHKeys(db.Model):
     __tablename__ = "ssh_keys"
+    id = db.Column(db.Integer, primary_key=True)
     user_id = db.Column(
-        db.Integer, db.ForeignKey("users.id", ondelete="CASCADE"), primary_key=True
+        db.Integer, db.ForeignKey("users.id", ondelete="CASCADE")
```
ConnorNelson, line 684: Let's also add an `index=True` to make sure we can quickly locate by `user_id`.

dojo_theme/static/js/dojo/settings.js
```diff
@@ -75,6 +75,7 @@ function button_fetch_and_show(name, endpoint, method,data, success_message, con
 
 $(() => {
     form_fetch_and_show("ssh-key", "/pwncollege_api/v1/ssh_key", "PATCH", "Your public key has been updated");
+    form_fetch_and_show("delete-ssh-key", "/pwncollege_api/v1/ssh_key/delete", "PATCH", "Your public key has been deleted");
```
ConnorNelson, line 78: Fix for `DELETE`.

dojo_theme/templates/settings.html
```diff
@@ -82,13 +82,37 @@ <h1>User Settings</h1>
           {% endwith %}
         </div>
 
+
         <div class="tab-pane fade" id="ssh-key" role="tabpanel">
+          <b><label for="enter-name">Current SSH Keys</label></b>
+          <small class="form-text text-muted">SSH Keys Currently Registered to this Account</small>
+          <br>
+
+          {% for key in ssh_key %}
```
ConnorNelson, line 91: Fix for `ssh_keys`.

ConnorNelson (changes requested):
Thanks for your contribution! I know some people have been asking for this for a while, so it's exciting to see this get added.

ConnorNelson:
Hey thanks for contributing this! I'm sorry I've been super busy finishing up the semester, but this is a top priority that I will get to this week! Part of the delay is confirming what any database migrations might look like as a result of database changes from this.

ConnorNelson:
Thanks for the PR! Made some minor changes, and merging now. Sorry it took so long before I could finally get to this!

Also as some brief documentation for the db migration (since we're currently doing db migrations manually....):
```sql
ALTER TABLE ssh_keys DROP FOREIGN KEY `ssh_keys_ibfk_1`;
ALTER TABLE ssh_keys DROP PRIMARY KEY;
ALTER TABLE ssh_keys MODIFY value VARCHAR(512) NOT NULL;
ALTER TABLE ssh_keys ADD PRIMARY KEY (user_id, value);
ALTER TABLE ssh_keys ADD CONSTRAINT `ssh_keys_ibfk_1` FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE;
```

PR #399, opened 2024-04-15
Add open slides in new window button

f4-u57:
Use case: open lecture slides on different tabs instead of viewing them in the related module page.

### DIFF
--- a/dojo_theme/templates/module.html
+++ b/dojo_theme/templates/module.html
@@ -44,6 +44,7 @@
           <div class="embed-responsive embed-responsive-16by9">
             <iframe class="embed-responsive-item" data-src="{{ src }}"></iframe>
           </div>
+          <button onclick="window.open('{{ src }}', '_blank')" class="btn btn-primary">Open Slides in New Window</button>
           {% endif %}
 
           {% elif resource.type == "markdown" %}
--- a/dojo_theme/templates/module.html
+++ b/dojo_theme/templates/module.html
@@ -44,7 +44,7 @@
           <div class="embed-responsive embed-responsive-16by9">
             <iframe class="embed-responsive-item" data-src="{{ src }}"></iframe>
           </div>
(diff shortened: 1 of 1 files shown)


Scoder12:
Just adding this:
In the firefox context menu, you can "this frame > open frame in new tab"
Chrome doesn't have an equivalent item, but you can "view frame source" then delete `view-source:` from the beginning of the URL. 

If you want, you can also strip `/embed` from the end of the URL to get the editor view. It comes down to personal preference but to me the toolbar being stuck at the bottom looks weird when the slides take up the whole viewport.

f4-u57:
I see. I've been just viewing source and getting the slides link there...

Yeah, I prefer the no `/embed` too because it shows preview of the slides in the editor view.

ConnorNelson:
What's the "real" goal here? You say to open it in a different tab; is the goal from there to open it full screen? To see the side bar to quickly find a slide?

I ask because I don't think I've ever wanted to do this, so it would be useful to understand *why* you want to.

I will say that I've been thinking about a pretty big interface overhaul, especially as it relates to a module page, so it would be useful to understand what sorts of workflows are missing.

f4-u57:
The goal is to open the slides in multiple tabs so I can see the topics in the tab names, then I preview the slides in the editor window (without the `/embed`) to go to the slide I think I need quickly. I find navigating the all the slides in the same module page a bit hard to use and there is no preview of slides. I also open lectures in multiple tabs for the same reason and with that I can just click on the yt lecture. I don't know maybe this is a rare use case.

zardus:
I think being able to open the slides in a new window is a great idea. I haven't checked out the look with this PR; could you upload a screenshot?

f4-u57:
[screenshot](https://i.imgur.com/3yAGplH.png)

ConnorNelson:
Yeah, honestly I don't necessarily love the aesthetics of this, but I'm definitely going to favor the utility over that for now. Thanks for the PR!

PR #404, opened 2024-04-26
Dojo: Support static and dynamic pages

ConnorNelson:


ConnorNelson:
I still need to add testcases.

PR #406, opened 2024-05-09
fix vulnerability reported by B0nk3rZ

zardus:
This checks that the shebang of the SUID script matches the runtime reality of the SUID interpreter launcher. Without this, very bad things (unintended privilege escalation within the user container) happen.


zardus:
Fix is already deployed on the central instance (https://pwn.college), but anyone running their own instance should upgrade ASAP.

PR #411, opened 2024-05-15
Do not show Profile page for admin since it always returns 404

mudongliang:


mudongliang:
?

ConnorNelson:
I’m not sure that I understand what issue this PR addresses. Admins should be able to view their own profile (maybe not if they have their status set to hidden)?

mudongliang:
I see. If admin is allowed to see its profile page, this PR is useless. Let's close it.

mudongliang:
Even if you set admin as visible, the profile page still shows 404 :/

PR #412, opened 2024-05-15
DM users upon successful discord linking

zardus:
This lays the groundwork for a hypothetical `Discord Dojo` with an actual flag challenge. Opening this PR for discussion.

Not sure how I feel about this versus just having the bot grant the White Belt role right upon linking. It's thematically nicer this way, but you can already see the potential for later phishing shenanigans once we inadvertently train users that we DM them from mysterious "pwn.college" accounts (as you can see from my warning in discord.py).

### DIFF
--- a/dojo_plugin/pages/discord.py
+++ b/dojo_plugin/pages/discord.py
@@ -69,4 +69,12 @@ def discord_redirect():
         db.session.rollback()
         return {"success": False, "error": "Discord user already in use"}, 400
 
+    # DM the user the flag for white belt
+    discord_dojo = Dojos.from_id("discord-dojo").first()
+    if discord_dojo:
+        discord_challenge == discord_dojo.modules[0].challenges[0]
+        flag = serialize_user_flag(user_id, discord_challenge.challenge_id)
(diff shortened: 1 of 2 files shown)


ConnorNelson:
It does sound kind of fun. My biggest worry honestly would be Discord being not thrilled with us bot-DMing users, wouldn't want the bot to get banned or something for spam. I think Discord would be more approving of a `/flag` command or something, that way it's a user-initiated message, but maybe I'm over thinking the Discord spam policy.

One simple way to pass the flag to the Discord bot (since it's command-processing is running unconnected to CTFd, etc) would be `/flag <TOKEN>`, where TOKEN is the flag encrypted and base64'd, and the secret is shared with the bot. Or we just share the flag-derivation logic and secret with the bot.

Alternatively though, maybe we don't want a challenge. Do you *have* to use Discord to "complete" pwn.college? I think you're probably a little silly if you don't, but I'm sure we'll have users that argue otherwise, and we don't necessarily need to require (or even strongly encourage via flags) that all users join the Discord.

Just some thoughts. I'm relatively neural.

PR #420, opened 2024-05-18
Get us out from under CTFd's routeocracy

zardus:
I'm currently trying to keep `/` of pwn.college updated with the white belt for the Getting Started dojo, and it's a disaster. This change removes CTFd's management of `/`, but it is still a WIP.

Basically: where do we want to keep the index? Some options:

1. Fully contained in the template as HTML (that's easy enough).
2. As a markdown loaded from elsewhere under dojo_theme?
3. Loaded from the welcome dojo's pages?

Alternatively, we can also redirect directly to the welcome dojo, with the welcome dojo (with an updated description and so on) just straight-up greeting visitors.


ConnorNelson:
Instead of `sed` patching, let's do monkey patching, like we already do, as seen in: https://github.com/pwncollege/dojo/blob/master/dojo_plugin/__init__.py#L116-L117

For reference, this is the current logic:
https://github.com/CTFd/CTFd/blob/3.6.0/CTFd/views.py#L389

We would define `app.view_functions["views.static_html"] = static_html`.

zardus:
Okay, I monkeypatched and basically moved /index.html to /dojo_theme/templates/index.html (with some de-markdowning applied). I figured we do this initially, since it's the lowest impact (versus figuring out a general place to retrieve page content from). Now we can sanely update index.html!

This is ready to squash-merge.

PR #422, opened 2024-05-19
Resolve symlinks in a subdir of challenge dir

etrickel:
Modified DojoChallenges.challenge_paths to return an absolute and relative path to support 1 level deeper symlinks.

The challenge_paths method is used by insert_challenge, which now creates the archive path using /challenge and the relative_path returned by challenge_paths.

### DIFF
--- a/challenge/Dockerfile
+++ b/challenge/Dockerfile
@@ -511,8 +511,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         gdb-multiarch
         gedit
         genisoimage
-        gprolog
         gnupg-utils
+        gprolog
         hexedit
         icdiff
         iproute2
--- a/challenge/Dockerfile
+++ b/challenge/Dockerfile
@@ -102,7 +102,7 @@ RUN <<EOF
     code-server --extensions-dir=/opt/code-server/extensions \
         --install-extension ms-python.python \
         --install-extension /tmp/ms-vscode-cpptools.vsix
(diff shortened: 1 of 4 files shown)


ConnorNelson:
Can you give an example of what this changes so I can better understand the goal here?

etrickel:
For the programming assignments, I have multiple levels that share tests, I don't want all the tests to show up in the /challenge directory, but would prefer them to show up in the `/challenge/tests` folder

If I create a symlink in the root of the challenge directory (e.g., `06-mud/level-06-01/.init -> ../.init`) it works just fine because the .resolve() in python is used at that level to resolve the path of the symlink. However, the resolve is not run on deeper directories. As a result, if I have a symlink in `06-mud/level-06-01/tests`
image
which resolves to `06-mud/tests/test1.1.json`

It will not work in the spawned challenge because it just copies the symlink file. 
From the challenge container, it looks like this
image

It's red because from `/challenge/tests` there is no `../../tests` in the container.

The code I added for the PR performs the resolution 1 layer deeper, so that the challenges can have symlinks in the gitrepo in both the challenge folder and its immediate child folders. 

I hope that helps.

etrickel:
I went ahead and combined the addition of some additional tools I'll need for the course:
jq - function based JSON query tool
racket - scheme language support
gprolog - prolog language support
sqlite3 - sqlite querying tool for SQL support

etrickel:
Added a bugfix for cpptools debugger. The fix sets 3 more files to executable to match a standard installation. 
The files that I added to the `chmod` are :
- /debugAdapters/bin/OpenDebugAD7
- /LLVM/bin/clang-format
- /LLVM/bin/clang-tidy

PR #436, opened 2024-06-01
improve error message when repository is left blank

Scoder12:
Before, we would just fall through both if-conditions, resulting in "dojo_dir referenced before assignment"


ConnorNelson:
Happy to merge, just wondering how you hit this edge case, or did static analysis in your editor catch it?

Scoder12:
I was trying to add the example dojo to my local instance. Saw the placeholder text and just hit the button without thinking, lol.

PR #438, opened 2024-06-06
Create docker environment for using nix inside challenges

supercoolspy:


ConnorNelson:
I want:

```sh
docker build -t build-nix . &&  docker run -v ./nix:/nix/store build-nix
```

After:
- Inside `./nix-store` should be all of my "nix store".
- Inside `./nix-store` should also be `./nix-store/bin` which contains symlinks to any program I might care about (e.g. code-server, novnc, xvnc [tigervnc], fluxbox, start-stop-daemon).

Now, I should be able to run:
```
docker run -it --rm -v ./nix:/nix ubuntu:24.04

$ /nix/bin/code-server
...
$ /nix/bin/fluxbox
...
etc
```

And it should just work. We know it works DEFINITELY when we see code-server in the browser (running ubuntu:24.04 image, with just 1 volume mount) AND ALSO fluxbox.

PR #439, opened 2024-06-07
Implement async peer support "workspace tokens"

### LINKED ISSUE #419: Asynchronous Peer Support
Instructors and TAs want to be able to asynchronously support a student. We need to make it so instructors can start up a challenge as a student, and see that student's files (to quickly/easily debug them). Here's the workflow/technical details I imagine:

1. Student generates an access-token, which lasts for a user-specified amount of time (suppose some 1D default, 180D max, 1D min). They can see the list of all access-tokens they generated (including when they generated them, and how much longer the token will last).
2. Student is loudly informed that this access-token grants holders read-only access to their entire home directory (and the ability to start their challenge instance).
3. Student shares token with instructor (who might also share with TAs). Instructor might even tell all students at the start of the semester to generate a 180D token and share it right away.
4. Not clear what the UI for using this token is, but a simple option would be an `?access=TOKEN` in the URL. Having this causes `start` to start the challenge (still normal docker name of user_ID, where ID is instructor's user ID), but it forces the start logic to use the student's instance, and also mount the student's home directory read-only (make sure still nosuid) at something like `/home/other` (probably come up with a better name than "other"). Instructor still gets their normal home directory mounted in, and can access via all the normal access methods. We probably also shouldn't put a real flag in there (maybe practice-mode only, but some challenges don't allow practice mode now, idk).

### LINKED ISSUE #440: Remove `random_home_path`
Back in the world of crazy nginx redirects to unix sockets in the home directory, we were afraid of people doing crazy `../../../../../var/homes/nosuid/4242/.dojo/desktop.sock` sorts of symlinks to enter someone else's workspace. And so we implemented `random_home_path` (which I always internally called PSLR, path space layout randomization) in order to avoid this sort of symlink attack. 

We left unix sockets a while ago in favor of tcp. This means that `random_home_path` is no longer necessary. We should reduce DOJO complexity, and remove it. We should just have `/var/homes/data/USER_ID` which is a sparse 1GB ext4 disk image, and `/var/homes/data/nosuid/USER_ID` which is that ext4 image mounted nosuid. Obviously it would be sweet if docker supported this natively (mounting a volume from a disk image), but it doesn't (I don't think that has changed?) and so we will continue doing it ourselves.

### LINKED ISSUE #449: Invalid string escape
My editor just caught this while working on the code:

https://github.com/pwncollege/dojo/blob/3878ec0320c953d3e9f7d058f97318d7c79c04ec/dojo_plugin/api/v1/docker.py#L62

Shouldn't this be an `r"` string?

Scoder12:
Closes #419 
Closes #440 
Closes #449


ConnorNelson:
I briefly discussed this on Discord, but I am moving this discussion here.

I know I previously stated read-only, but let's instead be a little fancier and do an overlay (i.e. OverlayFS).

Suppose we have some users 17 and 4242 (bind mount instead of ext4 for example simplicity):
```
root@ubuntu-server:/tmp/test/homes# mkdir data nosuid
root@ubuntu-server:/tmp/test/homes# mkdir data/17 data/4242
root@ubuntu-server:/tmp/test/homes# mkdir nosuid/17 nosuid/4242
root@ubuntu-server:/tmp/test/homes# mount --bind -o nosuid data/17 nosuid/17
root@ubuntu-server:/tmp/test/homes# mount --bind -o nosuid data/4242 nosuid/4242
```

Now suppose 17 is granted access to 4242, and 17 starts up an access workspace for 4242.  We don't want 17 to be able to modify 4242's home directory (hence the original read-only proposal), but in fact read-only is not the only way to facilitate this. We can instead grant 17 access to an *overlay* of 4242's home, in which 4242 will never seen any modifications that 17 might make (this is how docker *containers* are implemented on top of docker *images*). This requires a lowerdir (what is being overlayed on top of), upperdir (where any changes are saved to), a workdir (overlayfs demands a place to work), and then the overlay (where this overlay effect is made visible to, i.e. some overlay of upperdir over lowerdir). Probably, 17 should have a unique overlay of 4242:
```
root@ubuntu-server:/tmp/test/homes# mkdir overlay
root@ubuntu-server:/tmp/test/homes# mkdir -p overlay/17/4242/overlay overlay/17/4242/upperdir overlay/17/4242/workdir
root@ubuntu-server:/tmp/test/homes# mount -t overlay overlay -o lowerdir=nosuid/4242,upperdir=overlay/17/4242/upperdir,workdir=overlay/17/4242/workdir,nosuid overlay/17/4242/overlay
root@ubuntu-server:/tmp/test/homes# findmnt overlay/17/4242/overlay
TARGET                                  SOURCE  FSTYPE  OPTIONS
/tmp/test/homes/overlay/17/4242/overlay overlay overlay rw,nosuid,relatime,lowerdir=nosuid/4242,upperdir=overlay/17/4242/upperdir,workdir=overlay/17/4242/workdir
```

Now we can start the access workspace:
```
root@ubuntu-server:/tmp/test/homes# docker run -it --rm -v ./overlay/17/4242/overlay:/home/hacker -v ./nosuid/17:/home/me ubuntu:20.04
root@4b950d5c065d:/# findmnt /home/hacker
TARGET       SOURCE  FSTYPE  OPTIONS
/home/hacker overlay overlay rw,nosuid,relatime,lowerdir=nosuid/4242,upperdir=overlay/17/4242/upperdir,workdir=overlay/17/4242/workdir
root@4b950d5c065d:/# findmnt /home/me
TARGET   SOURCE                             FSTYPE OPTIONS
/home/me /dev/sdc2[/tmp/test/homes/data/17] ext4   rw,nosuid,relatime
```

In this case I've also opted to place 17's overlay of 4242 at /home/hacker and 17's own home at /home/me. Probably this is nice so that it's an even-closer replica of 17's workspace for that challenge (e.g. maybe the challenge wants to read from /home/hacker, or even modify something in /home/hacker, etc). This also mean's the supporter will see the same effect's automatically of /home/hacker/.bashrc, etc. It's about as identical as we can get (technically the environment could detect that /home/hacker is now an overlay, and that /home/me exists, but whatever). This also means that some super-invested in debugging someone else's solution can easily just modify /home/hacker/x.py without having to copy it and account for any possible crazy implications of that.

This also has nice security properties. My testing finds that writes to a named pipe will not follow through to the lowerdir, whereas writes to a named pipe in a read-only mount in fact *do* pass through. We ideally want to avoid the accessee's ability to in any way leak data from an access workspace.

ConnorNelson:
I think while you're messing around with home mounting anyways, it would also be great if you resolved #440 with this PR.

dojo_plugin/api/v1/docker.py
```diff
@@ -5,25 +5,42 @@
...
 
+logger = logging.getLogger(__name__)
+
 docker_namespace = Namespace(
     "docker", description="Endpoint to manage docker containers"
 )
 
 
-def start_challenge(user, dojo_challenge, practice):
+def start_challenge(owner, challenged_user, dojo_challenge, practice, partner=None):
```
ConnorNelson, line 37: Lets use this interface:
```py
def start_challenge(user, dojo_challenge, practice, *, as_user=None):
    as_user = as_user or user
```

A user starts a challenge, and they might start that challenge "as" another user (by default it's "as" themself).

Basically everything will use `user`, except:
- Mount `/home/hacker` from `as_user`
- Mount `/home/me` from `user`, but only mounted if `user != as_user` else no mount.
- Pass `as_user` to `insert_challenge` instead of `user`.

dojo_plugin/models/__init__.py
```diff
@@ -708,6 +708,31 @@ class Emojis(Awards):
...
+class WorkspaceTokens(db.Model):
+    __tablename__ = "workspace_tokens"
+    id = db.Column(db.Integer, primary_key=True)
+    user_id = db.Column(db.Integer, db.ForeignKey("users.id", ondelete="CASCADE"))
+    created = db.Column(db.DateTime, default=datetime.datetime.utcnow)
+    expiration = db.Column(
+        db.DateTime,
+        default=lambda: datetime.datetime.utcnow() + datetime.timedelta(days=30),
+    )
+    value = db.Column(db.String(128), unique=True)
```
ConnorNelson, line 725: Instead of a generic value, we should probably have a `authorizee_user_id` (and maybe rename `user_id` to `authorizer_user_id`?) This should also be a ForeignKey to a users.id, delete on cascade, etc.

ConnorNelson: Also, I absolutely agree with your thinking around not subtyping the already-existing CTFd tokens since it's baked into security-critical paths.

ConnorNelson:
One final thought for now while I'm thinking about this. What are you thinking about as the UX for granting access?

For example, I think we probably don't want someone to be able to spam grant access to a bunch of user's without them wanting that access. Maybe a shareable url, the accessee clicks it, hits some yes box, done? 

Then the other side is how do I actually start the challenge as a user I have access to? Maybe this is also through the URL and I'm mistaken about my earlier comment of an`authorizee_user_id`; we just want a random value that theoretically anyone can use (don't share the URL/value publicly...). That random value can just show up in a url and suddenly it passes through the start challenge button? Curious if you have thoughts on this! No matter what, we should make sure it is *extremely* clear what this mechanism does (don't want any user confused that this effectively grants access to the entire content's of their home directory).

dojo_plugin/models/__init__.py
Scoder12: > Instead of a generic value
the value field holds the token itself. are you saying we should replace sharing a token string with specifying the username you want to authorize, kinda like github repo invitations? or are you saying we should we keep the token but limit it to a certain user?

ConnorNelson: To further clarify, I was initially envisioning sharing with a specific user, but I think the UX is much simpler by embedding that "authorization" into the URL itself (i.e. if you have the URL, you *are* authorized) as I realized in my most recent comment.

dojo_plugin/api/v1/docker.py
```diff
@@ -327,16 +326,16 @@ def start_challenge(user, dojo_challenge, practice, *, as_user=None):
...
-
     umount_existing_overlay(user)
-    setup_home(user)
-    mounts = [("/home/hacker", HOST_HOMES / "nosuid" / str(user.id))]
+
+    setup_home(as_user)
+    mounts = [("/home/hacker", HOST_HOMES_NOSUID / str(as_user.id))]
     if as_user != user:
-        setup_home(as_user)
+        setup_home(user)
```
Scoder12: Just curious, why did you end up swapping user and as_user here? If I'm understanding correctly either order works so it shouldn't matter.

dojo_plugin/api/v1/docker.py
```diff
@@ -337,7 +337,7 @@ def start_challenge(user, dojo_challenge, practice, *, as_user=None):
         setup_home(user)
         setup_user_overlay(user, as_user)
         mounts = [
-            ("/home/hacker", HOST_HOMES_OVERLAY / str(as_user.id) / "merged"),
+            ("/home/hacker", HOST_HOMES_OVERLAY / str(user.id) / "merged"),
             ("/home/me", HOST_HOMES_NOSUID / str(user.id)),
```
Scoder12, line 341: Why did you change them to be the same lol

ConnorNelson: In my head it was the `as_user`'s home (because it is), but since this is an *overlay* of course as you originally had `user` is correct.

dojo_plugin/api/v1/docker.py
ConnorNelson: It doesn't technically make a difference as you point out. But in the same way that `insert_challenge` takes an `as_user`, the "`as_user`"  has the "priority", by default when we're setting up a home, it's the `as_user`'s home, and then if there's also a separate `user`, we need to set up for them also.

PR #441, opened 2024-06-14
Add SageMath to Dockerfile

prathamgupta36:
This PR adds SageMath to the Dockerfile. SageMath is a powerful mathematics software system that can be useful for various mathematical computations and research. The installation includes all necessary dependencies and builds SageMath from source.

### DIFF
--- a/challenge/Dockerfile
+++ b/challenge/Dockerfile
@@ -542,6 +542,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         php-mysql
         psutils
         python3-ipdb
+        python3-sagemath
         qemu-system-x86
         qemu-user
         qemu-utils


ConnorNelson:
How did you decide the package is named "python3-sagemath"? It seems to be named "sagemath".

prathamgupta36:
I tried using just `sagemath` but I still get `bash: sage: command not found`, someone on stackoverflow suggested this might be viable but since I don't know how to test it in `dojo`, I used that. I also found this after I committed- https://packages.ubuntu.com/jammy/amd64/python/python3-sage.

prathamgupta36:
For `sagemath` to work, I had to use `sudo apt-get update` and then I was able to use `sudo apt-get install sagemath`

ConnorNelson:
Yeah, I haven't merged this PR yet, so there definitely won't be sagemath on production. That package exists in 22.04, not 20.04 (which is the version we are on). Can you change this PR to be `sagemath`?

prathamgupta36:
Just changed it, would this work?

PR #443, opened 2024-06-15
adding a feature to sync with Canvas using a Canvas API token. Added …

etrickel:
Adding a feature to sync with Canvas using a Canvas API token. 
- Added an endpoint for admin initiated sync of the entire class admin/canvas_sync. 
- Added an automatic update that occurs on successful flag submission.

### DIFF
--- a/dojo_plugin/pages/canvas.py
+++ b/dojo_plugin/pages/canvas.py
@@ -36,19 +36,14 @@ This is called from the the DojoFlag class on a successful flag submission.
 It updates a single user's Canvas grade for the module that the the user just sucessfully submitted 
 """
 @authed_only
-def sync_challenge_to_canvas(challenge_id=None, user_id=None):
-        
-    if challenge_id is None or user_id is None:
-        return
-
+def sync_challenge_to_canvas(challenge_id, user_id):
     dojo_chal = DojoChallenges.query.filter(DojoChallenges.challenge_id == challenge_id).first()
-    dojo = Dojos.query.filter(Dojos.dojo_id == dojo_chal.dojo_id).first()
-    
-    log.info(f"Looking for {dojo.id=} {dojo_chal.module.id=} {user_id=}")
+    if dojo_chal:
+        dojo = Dojos.query.filter(Dojos.dojo_id == dojo_chal.dojo_id).first()
 
-    posting_results = do_canvas_sync(dojo, user_id=user_id, module_id=dojo_chal.module.id)
+        posting_results = do_canvas_sync(dojo, user_id=user_id, module_id=dojo_chal.module.id)
     
-    log.info(json.dumps(posting_results))
+        log.info(json.dumps(posting_results, indent=2))
 
 
 """ 
@@ -58,19 +53,15 @@ However, if both user_id and module_id are supplied it will do a single user/mod
 def do_canvas_sync(dojo, user_id=None, module_id=None):
     
     canvas_token = dojo.course.get("canvas_token","")
-    log.info(f"initial checks {canvas_token=} {user_id=} {module_id=}")
+    
     canvas_api_host = dojo.course.get("canvas_api_host","")
-    canvas_course_id = dojo.course.get("canvas_course_id","")
+    canvas_course_id = dojo.course.get("canvas_course_id",0)
     if len(canvas_token) == 0:
         return {"status": "completed", "message": "Missing canvas_token in course.xml"}
     if len(canvas_api_host) == 0:
         return {"status": "completed", "message": "Missing canvas_api_host in course.xml"}
-    if len(canvas_course_id) == 0:
+    if canvas_course_id == 0:
         return {"status": "completed", "message": "Missing canvas_course_id in course.xml"}
-    
-    # non-admins may only perform single user / module update to canvas
-    if (user_id is None or module_id is None) and not dojo.is_admin():
-        abort(403)
     
     canvas_grade_data = {}
     
@@ -79,6 +70,8 @@ def do_canvas_sync(dojo, user_id=None, module_id=None):
     students = {student.user_id: student.token for student in dojo.students}
     course_students = dojo.course.get("students", [])
     missing_students = list(set(course_students) - set(students.values()))
+    
+    # if user_id is None, it's a bulk grade submission, get all users, else just get user_id
     if user_id is None: 
         users = (
                 Users
@@ -87,8 +80,8 @@ def do_canvas_sync(dojo, user_id=None, module_id=None):
                 .filter(DojoStudents.dojo == dojo,
                         DojoStudents.token.in_(course_students))
             )
-    elif user_id is not None:
-        # get the user 
+    else:
+        # get the user_id
         users = (
                 Users
                 .query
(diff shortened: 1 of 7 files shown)


dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+However, if both user_id and module_id are supplied it will do a single user/module update 
+"""
+def do_canvas_sync(dojo, user_id=None, module_id=None):
+    
+    canvas_token = dojo.course.get("canvas_token","")
+    
+    canvas_api_host = dojo.course.get("canvas_api_host","")
+    canvas_course_id = dojo.course.get("canvas_course_id",0)
+    if len(canvas_token) == 0:
+        return {"status": "completed", "message": "Missing canvas_token in course.xml"}
```
adamdoupe, line 60: What is a course.xml? I thought we used yaml?

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+        if "canvas_assignment_id" not in assessment:
+            continue
+        assessment_student_counter[assessment['id']] = 0
+        # if we have a value module_id, check for it and skip all the others
+        if module_id is not None and module_id != assessment["id"]:                
+            continue 
+                    
+        canvas_assignment_id = assessment["canvas_assignment_id"]
+        for grade_res in grades:                
+            grade_credit_percent = f"{grade_res['assessment_grades'][aindex]['credit'] * 100:.2f}%"
```
adamdoupe, line 114: How does the `.2f` map with what the student sees on the grade page in pwn.college? I'm just curious / worried about rounding issues and any discrepancies between what's on pwn.college and what's in canvas will cause problems.

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+        posting_results.append(assessment_student_counter)
+
+    return posting_results
+
+
+def post_grade_to_canvas(canvas_token, canvas_api_host, canvas_course_id, assignment_id, student_id, students_grade):
+    url = f"https://{canvas_api_host}/api/v1/courses/{canvas_course_id}/assignments/{assignment_id}/submissions/sis_user_id:{student_id}"
+    payload = {'submission': {'posted_grade': students_grade}}
+    
+    header = {'Authorization': 'Bearer ' + canvas_token}
```
adamdoupe, line 153: I'm not sold, but maybe an f-string here rather than string concatenation would be more clear (for `Bearer`)

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+    
+    r = requests.put(url, headers=header, json=payload)
+    
+    if r.status_code == 200:
+        return {"status": "success", "student_id": student_id, "json": r.json()}
+    else:
+        return {"student_id": student_id, "code": r.status_code, "text": r.text, "url": url, "canvas_token": canvas_token}
```
adamdoupe, line 160: This isn't setting `status` to `"fail"` or something like that.

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+
+def post_grade_to_canvas(canvas_token, canvas_api_host, canvas_course_id, assignment_id, student_id, students_grade):
+    url = f"https://{canvas_api_host}/api/v1/courses/{canvas_course_id}/assignments/{assignment_id}/submissions/sis_user_id:{student_id}"
+    payload = {'submission': {'posted_grade': students_grade}}
+    
+    header = {'Authorization': 'Bearer ' + canvas_token}
+    
+    r = requests.put(url, headers=header, json=payload)
+    
+    if r.status_code == 200:
```
adamdoupe, line 157: Are we 100% sure that this means success for Canvas? Maybe we can put a link to the docs above this line so that people can check in the future. 

I've seen horrible APIs that return 200 on failure and put the status in a JSON response.

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+    else:
+        return {"student_id": student_id, "code": r.status_code, "text": r.text, "url": url, "canvas_token": canvas_token}
+    
+
+def post_bulk_grade_data_to_canvas(canvas_token, canvas_api_host, canvas_course_id, assignment_id, canvas_grade_data):
+    url = f"https://{canvas_api_host}/api/v1/courses/{canvas_course_id}/assignments/{assignment_id}/submissions/update_grades"
+    payload = {'grade_data': canvas_grade_data}
+    
+    header = {
+                'Authorization': 'Bearer ' + canvas_token,
```
adamdoupe, line 168: Same as other `Bearer` comment

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+                "Content-Type": "application/json"
+            }
+    
+    r = requests.post(url, headers=header, json=payload)
+    
+    r_json = r.json()
+    
+    if r.status_code == 200:
```
adamdoupe, line 176: The logic in the following doesn't mirror `post_grade_to_canvas`, is there a reason for this? If not, let's have them both match.

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+        return {"student_id": student_id, "code": r.status_code, "text": r.text, "url": url, "canvas_token": canvas_token}
+    
+
+def post_bulk_grade_data_to_canvas(canvas_token, canvas_api_host, canvas_course_id, assignment_id, canvas_grade_data):
+    url = f"https://{canvas_api_host}/api/v1/courses/{canvas_course_id}/assignments/{assignment_id}/submissions/update_grades"
+    payload = {'grade_data': canvas_grade_data}
+    
+    header = {
+                'Authorization': 'Bearer ' + canvas_token,
+                "Content-Type": "application/json"
```
adamdoupe, line 169: Line 153 doesn't have this, is it required there? If so, make sure both match.
https://github.com/pwncollege/dojo/pull/443/files#diff-d8c31178557f647f9c37f7aacc0ca7ec179bfde39f68f0aca5ada18b6ee96f1aR153

dojo_plugin/pages/canvas.py
```diff
@@ -0,0 +1,225 @@
...
+    headers = {
+        "Authorization": f"Bearer {access_token}",
+        "Content-Type": "application/json"
+    }
+    completed = []
+    check_count = 0
+    max_checks = 10
+    most_recent = dict()
+    while len(completed) < len(progress_urls):
+
```
adamdoupe, line 197: blank line

dojo_plugin/pages/canvas.py
etrickel: The display for admins and students use the templates grades_admin.html and course.html they both use %2f

dojo_plugin/pages/canvas.py
etrickel: I'm not sure, they are both clear to me, but I'll change them all to be f-strings.

dojo_plugin/pages/canvas.py
etrickel: It's a mistype, I'll fix.

dojo_plugin/pages/canvas.py
etrickel: All I found is that the API will return 200, 201, or 204 for success depending on the endpoint and 4XX and 5XX for errors. 
I'll update the check to check for 200, 201, and 204. But I could not verify that was the case for this specific endpoint. In my testing, I did see a few failures that reported as a failure. 

For the singular post grade to canvas, I wanted to return a message to the student confirming, but the plugin interfaces with the UI via the `DojoFlag.compare` method, which only returns a true/false and the and the UI creates the success message. 

Fortunately, the grade will still get updated via a subsequent bulk update or another flag submission within the same module.

dojo_plugin/pages/canvas.py
etrickel: I'm going to create the header in the calling method that looks up the token, they should all be the same.

PR #444, opened 2024-06-17
Workspace: Handle service starts gracefully

ConnorNelson:


ConnorNelson:
I think it would make sense to do something like use javascript to poll every second to see if the service is ready (e.g. container exists, container is running, service is running). Once that has been confirmed, let the iframe show up. In the meantime have a "loading" animation. Do this for 20s and then give up and tell the user no challenge is running.

Scoder12:
does that mean that if you truly don't have a challenge running, you have to wait 20s to find out?

ConnorNelson:
Yeah it definitely would.

What we could do is be more informative about the process as we are with SSH. Maybe instead we check if the container exists in any state at all (e.g. even just "Created"); if not, immediately report no challenge started, otherwise poll for service-readiness for ~20 seconds.

ConnorNelson:
#478 has a better discussion of what this should look like.

PR #446, opened 2024-06-18
Workspace: Build and mount nix

ConnorNelson:
Based on work in #438.


supercoolspy:
```bash
$ mkdir -p /run/current-system/sw/bin/ && ln -s $(which dbus-daemon) /run/current-system/sw/bin/dbus-daemon
$ mkdir -p /etc/dbus-1
# Remove infinite recursion line
$ cat $(dirname $(dirname $(which dbus-daemon)))/share/dbus-1/session.conf | grep -v '<include ignore_missing="yes">/etc/dbus-1/session.conf</include>' > /etc/dbus-1/session.conf
$ cp $(dirname $(dirname $(which dbus-daemon)))/share/dbus-1/system.conf /etc/dbus-1/system.conf
```

PR #451, opened 2024-07-01
Add a share-by-link solves csv/json endpoint

zardus:
Changing the solves endpoint to be link-shareable like the update endpoint for better integration. For the secret, I'm using a hash of the private key for the dojo's deploy key (didn't want to reuse the update code because granting update permissions shouldn't necessarily be the same as granting solves access).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:24.04
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
 ENV LC_CTYPE=C.UTF-8
 
 RUN apt-get update && \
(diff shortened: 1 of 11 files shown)


ConnorNelson:
https://docs.docker.com/reference/dockerfile/#env
```
The ENV instruction also allows an alternative syntax ENV <key> <value>, omitting the =.
...
The alternative syntax is supported for backward compatibility, but discouraged for the reasons outlined above, and may be removed in a future release.
```
What makes this new syntax more "modern"?

zardus:
Docker on my desktop was actively refusing to build with the "old" format... I figured we should just move with the times.

Do you like the rest of the changes? Itching to merge given the huge code move (to avoid conflicts).

ConnorNelson:
Just wanted to state that this PR looks good to me and I agree with the feature idea.

I think long term we might want to not represent this secret as a hash of the private key, but this seems pretty inconsequential for now (I do like the idea of deriving sub-secrets from a singular dojo secret). Also, in this PR you still keep `ENV LC_CTYPE=C.UTF-8`, so I'm a little bit confused about what is going on with your docker.

PR #455, opened 2024-07-04
Fix CTFd Healthcheck

### LINKED ISSUE #150: Fix CTFd Healthcheck
It's hitting `localhost:8000`, but on production, that doesn't hit us.

Scoder12:
Closes #150

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -98,7 +98,9 @@ services:
         - "-c"
         - |
           import requests
-          response = requests.get('http://localhost:8000')
+          response = requests.get(
+            'http://localhost:8000' if '${DOJO_ENV}' == "development" else "https://${LETSENCRYPT_HOST}"
+          )
           response.raise_for_status()
       interval: 10s
       timeout: 10s


Scoder12:
Proof the expansion is working properly: 
```console
root@ctfd:/opt/CTFd# while true; do ps aux | grep -F 'python -c' | grep -v grep; sleep 0.1; done
root         878  3.0  0.0   8764  4352 ?        Rs   19:29   0:00 python -c import requests response = requests.get(   'http://localhost:8000' if 'development' == "development" else "https://localhost.pwn.college
" ) response.raise_for_status()
```
(Disclaimer: you have to `apt update && apt install procps` for this snippet to work)

ConnorNelson:
Thanks! It looks like this isn't necessary anymore though; I'm not quite sure what has changed since that issue at this point but it seems like it has been resolved (as per production):
```
542fe54c1ba1   pwncollege-ctfd                     "/bin/sh -c 'if [ \"p…"   2 days ago           Up 2 days (healthy)    8000/tcp                                   ctfd
```

It seems like I may need to go through our backlog of issues and see what in fact is still open.

Issue #445, opened 2024-06-18
Make it harder to delete dojos by accident


zardus:
We have an awesome way of deleting dojos from the web site thanks to https://github.com/pwncollege/dojo/pull/346, but it's not sufficiently difficult to accidentally delete a dojo. We should have something similar to github's "enter the repository name to confirm" to make things safer.

zardus:
Relevant code: https://github.com/pwncollege/dojo/blob/master/dojo_theme/static/js/dojo/settings.js#L90

abhinavk454:
I want to work with this if it's open to OS contributors.

zardus:
Absolutely open, go for it!

JensHouses:
Added pull request #460 for this issue 🙂

JensHouses:
@zardus this issue can be closed 👍

zardus:
Good catch!

PR #460, opened 2024-07-12
JS: Improved dojo deletion confirmation

JensHouses:
- Updated button_fetch_and_show to include an abort message when user cancels action
- Improved confirmation logic for dojo-delete button

This change makes it harder to delete dojos by requiring confirmation and providing clear feedback when the action is canceled. This PR adresses #445 

Confirmation:
grafik
Cancel msg:
grafik


zardus:
Thanks!

PR #466, opened 2024-07-18
Make "Modules" singular if there is only one module in the dojo.

sriske2:
before:
ARM Dojo

after:
ARM Dojo


ConnorNelson:
Thanks!

Issue #105, opened 2022-10-27
QoL improvement: submit/next/prev buttons on navbar


zardus:
Aside from an internal command line tool, we could have small buttons on the navbar to pop up a modal to submit a flag, to launch the next challenge, and to launch the previous challenge.

JensHouses:
Hey everyone,

I noticed this issue has been open for a while. I'd love to work on it.

zardus:
Sure, we'd love the help!

One issue is that we're running out of space in the nav bar. However, we have this little dude:

![image](https://github.com/user-attachments/assets/2cf63b00-9b1e-4be4-a425-7170c752e39e)

We could replace it with a hamburger (https://fontawesome.com/icons/bars?s=solid) and have stuff like:

- `Hide Navbar` (or `Show Navbar` if Navbar is hidden)
- If challenge is started: `Submit Flag` `Restart Challenge` `Next Challenge` `Previous Challenge`
- If challenge is not started: `Start Recent Challenge`
- Maybe move `Admin` over here?

JensHouses:
Sounds like a good idea.
I see a solution for all points.
But I need some help for the start recent challenge:

How do I determine the recent challenge? 
Iterate through all unsolved  dojos -> modules -> challenges ? They are sorted by index I guess ?

zardus:
Hmm, good point… for now, let’s ignore the issue and just not have that one.On Jul 16, 2024, at 11:40 PM, Jens Häuser ***@***.***> wrote:﻿
Sounds like a good idea.
I see a solution for all points.
But I need some help for the start recent challenge:
How do I determine the recent challenge?
Iterate through all unsolved  dojos -> modules -> challenges ? They are sorted by index I guess ?

JensHouses:
![grafik](https://github.com/user-attachments/assets/4e9adb76-37ca-4340-be8e-78ba134d1a33)
What do you think about that layout ?

ConnorNelson:
What about a button in the middle (https://fontawesome.com/icons/rotate-left?f=classic&s=solid) with the current challege name that says "Restart `current`". 

And then maybe just something that more closely resembles the existing flag submit line at the bottom (we wouldn't need to say Mars anymore if its on the restart button):

<img width="1085" alt="image" src="https://github.com/user-attachments/assets/cd3242a3-c4c6-48b9-aa39-d7a31cd3ffc6">

Or something closer to that (fine if we have a slightly different feel to match this popup).

[...]

PR #473, opened 2024-07-22
 QoL improvement: submit/next/prev buttons on navbar

JensHouses:
solving #105

- The Dojo name is a link to the Dojo
- The Module name is a link to the Module
- Starting a new Challenge in the control panel updates the Highlighting of the module Page if open
- Starting a new Challenge in the control panel updates the control panel 
- Starting a new Challange in the module page updates the control panel
- if first or last challenge is reached the corresponding buttons gets invisible to keep the layout
- you can submit the flag to the current challenge in the control panel
- notifications are sticked to the top to overlay the menu

### DIFF
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -71,11 +71,6 @@ def active_module():
     if active:
     #return a json of the active challenge
         return {
-            "next_as_json": json.dumps(123),
-            "module_name": active.module.name,
-            "dojo_name": active.dojo.name,
-            "challenge_id": active.challenge_id,
-            "challenge_name": active.name,
             "c_previous": {
                 "module_name": challs['previous'].module.name ,
                 "module_id": challs['previous'].module.id,
@@ -105,7 +100,6 @@ def active_module():
             } if challs['next'] is not None else {},
         }
     return {}
-    #return active.challenge_id or "None"
 
 
 @dojo.route("/dojo/<dojo>")
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -16,7 +16,7 @@ from CTFd.cache import cache
 
 from ..utils import render_markdown
 from ..utils.stats import container_stats, dojo_stats
-from ..utils.dojo import (dojo_route, get_current_dojo_challenge, get_prev_cur_next_dojo_challenge, dojo_update, dojo_admins_only)
+from ..utils.dojo import dojo_route, get_current_dojo_challenge, get_prev_cur_next_dojo_challenge, dojo_update, dojo_admins_only
 from ..models import Dojos, DojoUsers, DojoStudents, DojoModules, DojoMembers, DojoChallenges
 
 dojo = Blueprint("pwncollege_dojo", __name__)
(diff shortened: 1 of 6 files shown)


dojo_plugin/pages/dojo.py
```diff
@@ -59,6 +59,49 @@ def view_dojo_path(dojo, path):
     else:
         abort(404)
 
+
```
zardus, line 62: Could you add a test case for this? Only a matter of time until we break this endpoint otherwise :crying_cat_face:

dojo_theme/templates/components/navbar.html
```diff
@@ -15,6 +15,90 @@
     </button>
     <div class="collapse navbar-collapse" id="base-navbars">
       <ul class="navbar-nav mr-auto">
+        <li id="challengeMenuButton" class="d-none nav-item dropdown">
```
zardus, line 18: What do you think about moving this to the right hand side, and making it always-condensed (like the logout button?). On a note-to-self, we should also always-condense the Profile and Settings probably, to keep the navbar from completely filling up. As things are right now, that navbar is super cluttered.

zardus: This might be a cool result:
image

zardus (changes requested):
This is AWESOME! So smooth to work through the chals. Thank you.

Just a few minor comments, and then I think we'll be ready to merge!!

dojo_plugin/pages/dojo.py
JensHouses: I'll take a look at the other tests and add a testcase 😄

dojo_theme/templates/components/navbar.html
JensHouses: Yeah I also thought about the flag symbol only.
As it might (hopefully) become a tool that people use more often then the dojo link I would rather put it to the left even without the label.
But I'm also fine with it being on the right side.

I really like that little panel 😁

JensHouses: grafik
this is result now

dojo_plugin/pages/dojo.py
JensHouses: grafik
added

zardus:
I'm about to take off on a flight and the internet is no good, so I can't pull this down to take a look locally, so we're going to just yolo it into production and take a look there!

Thank you for this contribution!! Huge QoL++

adamdoupe:
@zardus this should not have been merged without fixing/addressing https://github.com/pwncollege/dojo/issues/478

The next button doesn't really work, and this is honestly more frustrating to use: because it is really very nice, you want it to work and it's a major bummer when it doesn't.

JensHouses:
> @zardus this should not have been merged without fixing/addressing #478
> 
> The next button doesn't really work, and this is honestly more frustrating to use: because it is really very nice, you want it to work and it's a major bummer when it doesn't.

i added a PR to fix this #482

zardus:
> @zardus this should not have been merged without fixing/addressing #478
> 
> The next button doesn't really work, and this is honestly more frustrating to use: because it is really very nice, you want it to work and it's a major bummer when it doesn't.

But see, within 3 hours of yolo-merging, suddenly everything works perfectly ;-)

PR #479, opened 2024-07-23
Add ldd wrapper

### LINKED ISSUE #471: ldd incorrectly lists libraries from nix store
This is particularly relevant for heap related challenges where libc version is relevant.

```
hacker@practice~dynamic-allocator-misuse~level7-0:~$ /usr/bin/ldd /challenge/babyheap_level7.0
	linux-vdso.so.1 (0x00007fff9e3d6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6955e96000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6956095000)
hacker@practice~dynamic-allocator-misuse~level7-0:~$
```

```
hacker@practice~dynamic-allocator-misuse~level7-0:~$ ldd /challenge/babyheap_level7.0
	linux-vdso.so.1 (0x00007ffdf55c3000)
	libc.so.6 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6 (0x00007f16b859c000)
	/lib64/ld-linux-x86-64.so.2 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2 (0x00007f16b878b000)
```

Scoder12:
Closes #471

### DIFF
--- a/workspace/flake.nix
+++ b/workspace/flake.nix
@@ -21,6 +21,21 @@
         code-service = import ./services/code.nix { inherit pkgs; };
         desktop-service = import ./services/desktop.nix { inherit pkgs; };
 
+        ldd-wrapper = (pkgs.writeShellScriptBin "ldd" ''
+            ldd=/usr/bin/ldd
+            for arg in "$@"; do
+              case "$arg" in
+                -*) ;;
+                *)
+                  case "$(readlink -f "$arg")" in
+                    /nix/store/*) ldd="${pkgs.glibc.bin}/bin/ldd" ;;
+                  esac
+                  ;;
+              esac
+            done
+            exec "$ldd" "$@"
+        '');
+
         additional = import ./additional/additional.nix { inherit pkgs; };
 
         corePackages = with pkgs; [
@@ -36,6 +51,7 @@
           gnused
           hostname
           iproute2
+          (lib.hiPrio ldd-wrapper)
           less
           man
           ncurses
@@ -43,23 +59,6 @@
           util-linux
           wget
           which
-          (lib.hiPrio (pkgs.writeShellScriptBin "ldd" ''
-            #!${pkgs.runtimeShell}
-            ldd=/usr/bin/ldd
-
-            for arg in "$@"; do
-              case "$arg" in
-                -*) ;;
-                *)
-                  case "$(readlink -f "$arg")" in
-                    /nix/store/*) ldd="${pkgs.glibc.bin}/bin/ldd" ;;
-                  esac
-                  ;;
-              esac
-            done
-
-            exec "$ldd" "$@"
-          ''))
 
           init
           ssh-entrypoint
--- a/workspace/flake.nix
+++ b/workspace/flake.nix
@@ -43,6 +43,23 @@
           util-linux
           wget
           which
+          (lib.hiPrio (pkgs.writeShellScriptBin "ldd" ''
+            #!${pkgs.runtimeShell}
+            ldd=/usr/bin/ldd
+
+            for arg in "$@"; do
+              case "$arg" in
+                -*) ;;
+                *)
+                  case "$(readlink -f "$arg")" in
+                    /nix/store/*) ldd="${pkgs.glibc.bin}/bin/ldd" ;;
+                  esac
+                  ;;
+              esac
+            done
+
+            exec "$ldd" "$@"
+          ''))
 
           init
           ssh-entrypoint
--- a/workspace/flake.nix
+++ b/workspace/flake.nix
@@ -7,21 +7,22 @@
 
   outputs = { self, nixpkgs }: {
     packages = {
-      x86_64-linux = let
-        pkgs = import nixpkgs {
-          system = "x86_64-linux";
-          config = {
-            allowUnfree = true;
+      x86_64-linux =
+        let
+          pkgs = import nixpkgs {
+            system = "x86_64-linux";
+            config = {
+              allowUnfree = true;
+            };
           };
-        };
 
-        init = import ./init.nix { inherit pkgs; };
-        ssh-entrypoint = import ./ssh-entrypoint.nix { inherit pkgs; };
-        service = import ./services/service.nix { inherit pkgs; };
-        code-service = import ./services/code.nix { inherit pkgs; };
-        desktop-service = import ./services/desktop.nix { inherit pkgs; };
+          init = import ./init.nix { inherit pkgs; };
+          ssh-entrypoint = import ./ssh-entrypoint.nix { inherit pkgs; };
+          service = import ./services/service.nix { inherit pkgs; };
+          code-service = import ./services/code.nix { inherit pkgs; };
+          desktop-service = import ./services/desktop.nix { inherit pkgs; };
 
-        ldd-wrapper = (pkgs.writeShellScriptBin "ldd" ''
+          ldd-wrapper = (pkgs.writeShellScriptBin "ldd" ''
             ldd=/usr/bin/ldd
             for arg in "$@"; do
               case "$arg" in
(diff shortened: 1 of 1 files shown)


Scoder12:
```console
hacker@hello~apple:~$ ldd /bin/sh
        linux-vdso.so.1 (0x00007ffe291f9000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8dd12e3000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f8dd1508000)
hacker@hello~apple:~$ ldd /run/current-system/sw/bin/which
        linux-vdso.so.1 (0x00007ffda33f4000)
        libc.so.6 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6 (0x00007f5b05a08000)
        /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/ld-linux-x86-64.so.2 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2 (0x00007f5b05bf7000)
hacker@hello~apple:~$ ldd --version
ldd (Ubuntu GLIBC 2.31-0ubuntu9.16) 2.31
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

Scoder12:
Still works if filename isn't the first arg: 
```console
hacker@hello~apple:~$ ldd -v -v /run/current-system/sw/bin/which
        linux-vdso.so.1 (0x00007ffc425b3000)
        libc.so.6 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6 (0x00007f32082b7000)
        /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/ld-linux-x86-64.so.2 => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2 (0x00007f32084a6000)

        Version information:
        /run/current-system/sw/bin/which:
                libc.so.6 (GLIBC_2.3) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.3.4) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.14) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.33) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.4) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.34) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6
        /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib/libc.so.6:
                ld-linux-x86-64.so.2 (GLIBC_2.2.5) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2
                ld-linux-x86-64.so.2 (GLIBC_2.3) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2
                ld-linux-x86-64.so.2 (GLIBC_PRIVATE) => /nix/store/dbcw19dshdwnxdv5q2g6wldj6syyvq7l-glibc-2.39-52/lib64/ld-linux-x86-64.so.2
```

ConnorNelson:
Cool! Can we not fully inline it, let's do `ldd = pkgs.writeShellScriptBin ...`, maybe between `desktop-service ` and `additional` as its own new block where we'll start putting patched standard-software and then include ldd a simple single line.

How concerned are we with this `lib.hiPrio` thing? Does it break if someone else defines ldd unless you include it? Does it normally favor the later defined thing? How does nix handle this. Additionally does it make sense to say that an "environment" has a `pkgs.runtimeShell`? Not that we need to address this here, but to better understand, should we be using this in many of the places that simply say `pkgs.bash` as the shebang (trying to better understand standard nix customs).

Scoder12:
Sure, I can define it separately. 

`lib.hiPrio` is needed because `glibc.bin` is already getting installed, and it has a `/bin/ldd`. So if you try to install a second package with a `/bin/ldd`, nix doesn't know which one to point the symlink at and errors out: 
```
workspace-builder  | error: builder for '/nix/store/vmzbc9br1l402x36yhc2q4xd5ijri6fz-dojo-workspace-core.drv' failed with exit code 25;
workspace-builder  |        last 1 log lines:
workspace-builder  |        > error: collision between `/nix/store/iijf1hqfraqidksmjiwss5npn8lypvc9-ldd/bin/ldd' and `/nix/store/z5gvbj7p7dr89k42ncrl2mlv5v3ymjlp-glibc-2.39-52-bin/bin/ldd'
workspace-builder  |        For full logs, run 'nix log /nix/store/vmzbc9br1l402x36yhc2q4xd5ijri6fz-dojo-workspace-core.drv'.
```

`lib.hiPrio` makes our `ldd` script higher priority so that the symlink will point to us instead. According to the manual and the sources, there's a `lowPrio` too, and a `setPrio` to set it to a numeric value explicitly. So it seems like the priority is more precise like niceness rather than blunt like `!important` in CSS. 

If someone else defines ldd, it will get silently overridden by us unless our `hiPrio` is removed or they `setPrio`. As far as I understand, "ties" are broken by erroring out. 

(1/)

ConnorNelson:
Ok sweet, yeah I've seen nix fail out because of conflicting files. Good to know that this prio mechanism is available for resolving that.

Scoder12:
> Additionally does it make sense to say that an "environment" has a pkgs.runtimeShell?

Not 100% sure what you mean, but I think that the `runtimeShell` is more tied to our instance of `nixpkgs` rather than our user environment. 

> should we be using this in many of the places that simply say pkgs.bash as the shebang (trying to better understand standard nix customs).

Most of my knowledge of "nix customs" comes from reading the code samples in the manual and some of the nixpkgs source code. The code samples that do stuff like this all use `#! ${runtimeShell}` so I just started using that without fully knowing the implications. 

`runtimeShell` is indeed just an alias to bash. It's not documented in the manual, but someone posted on the forum that using `runtimeShell` "makes it easier to pin that to a particular version if necessary, while using bash in those directly might cause problems on bumps, which had to be fixed individually." What I think this means is that using `runtimeShell` for shebangs is traditional because if you want to change what shell nixpkgs uses internally, you can override `runtimeShell` instead of overriding `bash` itself. `stdenv.shell` is a thing too but, I'm not sure what the distinction between the two is. 

To answer the question: we will probably never use the overriding thing but it's just a bit cleaner to use `runtimeShell` (and a bit shorter, because you don't have to tack on `/bin/bash` at the end). 

Additionally, upon reading the manual more closely, `writeShellScript` and `writeShellScriptBin` actually add this exact `runtimeShell` shebang for you, so the one I added is actually unnecessary.

Scoder12:
Pushed fixes.

zardus:
yolo

Issue #478, opened 2024-07-23
Better reconnection for nonvnc on challenge restart


zardus:
We generate a new VNC password when we restart a challenge, and as a result, novnc can't automatically reconnect across challenge starts. This annoyance will get more noticeable as we streamline the UX (e.g., https://github.com/pwncollege/dojo/pull/473). We should probably patch novnc to refresh the page when a reconnection fails (which will also fix the DOS issue).

ConnorNelson:
https://github.com/pwncollege/dojo/blob/89dbadaea108ef4da75006067f8cfde5e7acbbcb/workspace/services/desktop.nix#L7 should have already fixed DOS.

I think this actually makes sense as a "noVNC patch". Instead what we should do is put javascript on the top level page that decides where the iframe will point at. Somehow, we need the ability for that javascript to be informed that it should change the iframe location.

For example:
- I receive an event that a new workspace has started and initialized; this JS knows the user is interested in the Desktop (because we are on /workspace/desktop)
- Start a sweet spinning "loading" animation
- Tell the backend to start the desktop service
- When that has finished, kill the "loading" animation and set the iframe location

We should do the same thing for *all* top-level /workspace/* routes.

Scoder12:
I would lean towards SSE as it has way less overhead: it's a normal HTTP request that stays open and streams a text-based protocol. Websocket requires an upgrade handshake between the client and server and every message has to be XORed. 

I took a look at the CTFd notification logic. A couple important files are [the route the clients connect to](https://github.com/CTFd/CTFd/blob/8ef0cdd916eab4226956d9d2d70adeccdc89f441/CTFd/events/__init__.py) and [the events_manager classes](https://github.com/CTFd/CTFd/blob/8ef0cdd916eab4226956d9d2d70adeccdc89f441/CTFd/utils/events/__init__.py#L70). Here's my understanding of how it works. Redis has a built in pub/sub mechanism. The event manager's job is to spawn a gevent task (basically a thread) to subscribe to redis. When an HTTP request comes through to the SSE endpoint, it returns a generator which is how flask streams stuff. The generator hooks into the event manager, encodes the events in the SSE format, and forwards them to the client. 

It *seems* like this would lock up a flask worker, but I'd have to test to be sure. Maybe the only solution is to add more flask workers.

ConnorNelson:
I think I also lean towards SSE, especially since we don't have a need for bidirectional communication, and probably even more importantly, CTFd already has logic for this!

[...]

PR #482, opened 2024-07-25
added novnc iframe src updating

JensHouses:
Added a check for the Workspace Deskop and if found replacing the src of the url to get a smoother experience while using the control panel.


adamdoupe:
Tested that this works in prod. A bit janky, but great for now.

Fixes #478

PR #485, opened 2024-07-29
Workspace: Add suid-interpreter

ConnorNelson:


workspace/suid-interpreter.nix
```diff
@@ -0,0 +1,149 @@
+{ pkgs }:
+
+let
+  suidSource = pkgs.writeText "suid_interpreter.c" ''
```
Scoder12, line 4: Isn't this long enough (and security critical enough) to be it's own file, with syntax highlighting and all? I know it's kinda hard to fit into the file tree logically

ConnorNelson: Yeah you convinced me, a good location was the thing stopping me. For now I've added a `core` directory to make the structure a little better.

ConnorNelson:
The big thing blocking this merge is going to be making sure suid-python correctly picks up the dependencies for various challenges (e.g. unicorn/asteval/flask). Probably testing against building-a-web-server and embroyasm will be sufficient to make sure I've set it up correctly.

PR #486, opened 2024-07-29
Dojos: Fix pull for non-linear history

ConnorNelson:


ConnorNelson:
@zardus sanity check: this makes sense right?

Scoder12:
LGTM

PR #511, opened 2024-08-05
UI: Prevent lingering loading message from overflowing

sriske2:
The current loading message animation just adds dots to "Loading." until the challenge spins up successfully or an error message is shown. There is nothing stopping it from adding dots until it overflows past the UI. This often happens when the dojo is seeing heavy usage and challenge instances are slow to spin up. 

The updated code just adds up to five dots then starts back at "Loading." continuously.


zardus:
Thanks!

PR #513, opened 2024-08-05
Adding Extra Late Due Date for Courses

etrickel:
Added a secondary due date called extra late date, the purpose of the extra late date is to have a secondary more severe late penalty. For example, a late penalty might be 10% but the extra late penalty might be 50%.

If the courses.xml does not use the `extra_late_date` field then the column will not show up on the grades pages.

I also added some additional width configurations to the custom.css so that the web page will expand when being viewed on a higher resolution screen.

image


etrickel:
Updated to the current version

PR #517, opened 2024-08-06
Infrastructure: Add multi-node support

ConnorNelson:
Basic implementation of multi-node:
- Docker over TCP
- Nodes connected via wireguard
- Workspaces end up on node `user_id % N`
- Homes (small files) over NFS
- Moved home creation/mounting/overlaying into autofs, and out of site logic: just ask for it, and it exists
- Node 0 is the main node, it runs the site / db / sshd / etc; default behavior is to also run workspaces (for easy development), but production setup is to not
- Node N!=0 are the workspace nodes, they run workspacefs and the workspace (user) containers, also build challenge image and workspace-builder runs (nix)
- Main container is used for both main node and workspace nodes, automatically disable unneeded services on workspace nodes
- We will need to manually synchronize updates (e.g. go into each node and update workspace-builder / challenge).

Main node startup:
- `-p 51820:51820/udp`: This is what wireguard talks over. This will be internet-facing. We can firewall if we really wanted to, but that seems unnecessary because you need the secret to join.

Workspace node startup:
- `-e DOJO_HOST=pwn.college`: where is the main node to connect into for wireguard
- `-e WORKSPACE_KEY=...`: secret for connecting into wireguard
- `-e WORKSPACE_NODE=1`: maybe this is a bad pattern, but its certainly the simplest. nodes self declare their node id, the main node is 0.
- We don't need to `-p` anything.
- Still `-v` code and data
- Still `--privileged`

---

Upgrade info:
- Bring main node down, unmount all homes
- `e2fsck -f /data/homes/data/* && tune2fs -U random /data/homes/data/*`: In order to use nfs, we need to actually have all our homes have unique UUIDs, otherwise nfs will collide and be mega-sad.
-  YOLO

---

Sometimes NFS gets sad (I think when we restart the main node infra that is running the nfs server). Simple fix:

On every worker node:
Kill the user containers, they're probably doomed anyways if NFS restarts (maybe there's a way to make that not be true). Kill the autofs mount holding all the nfs homes, restart autofs.
```
docker kill $(docker ps | grep user_ | awk '{print $1}')
umount -l /data/workspace/homes/mounts
service autofs restart
```

On main node:
```
exportfs -ra
```

So far this seems to work. Hopefully we'll better understand this situation as time goes on.


ConnorNelson:
In order to make NFS actually work, we will need to give (and continue to give) unique UUIDs to all the homes:
```
e2fsck -f /data/homes/data/*
tune2fs -U random /data/homes/data/*
```

NFS uses the UUID to differentiate different mounts, and under the current paradigm all homes have the same UUID, leading to wild inode collisions.

zardus:
I thought we were exporting all of /data/home as a single NFS mount, and
doing the ext4 mount on the nodes?

ConnorNelson:
So I realized that workspace access tokens complicate this. If the two users aren't on the same node, we'd have to migrate user homes dynamically by correctly unmounting/remounting at the new node. Additionally, we'd have to remember to unmount all the user homes on a node before taking it offline. This isn't the end of the world, but I'd like to avoid complexity like this.

I'm going to mount the actual homes (not ext4s) over NFS as a first step. If the performance on this feels terrible, we'll investigate NFSing the ext4s as an optimization. We're gonna start by valuing simplicity though for now.

ConnorNelson:
When the main node goes down, here is how to (manually) fix NFS on the nodes:

On workspace node(s):
```
docker kill $(docker ps | grep user_ | awk '{print $1}')
umount -l /data/nfs/homes/nosuid
service autofs restart
```

On main node:
```
exportfs -ra
```

The issue is that we get `cannot open directory '/data/nfs/homes/nosuid/1': Stale file handle`. Maybe there's another way to resolve this?

ConnorNelson:
Lets make sure we document how to bring up new nodes.

PR #529, opened 2024-08-14
Fix `/<dojo>/modules` API endpoint

### LINKED ISSUE #527: `/<dojo>/modules` API endpoint doesn't work
Calling the endpoint results in Internal Server Error
For instance:
https://pwn.college/pwncollege_api/v1/dojo/fundamentals/modules

https://github.com/pwncollege/dojo/blob/59b31c64b010f3b4b3f1e70fefc77d832d1a8882/dojo_plugin/api/v1/dojo.py#L155-L166

fpotier:
Should close #527

### DIFF
--- a/dojo_plugin/models/__init__.py
+++ b/dojo_plugin/models/__init__.py
@@ -405,6 +405,14 @@ class DojoModules(db.Model):
             resource.resource_index = resource_index
         self._resources = value
 
+    @hybrid_method
+    def visible(self):
+        return len(self.visible_challenges()) > 0
+
+    @visible.expression
+    def visible(cls, when=None):
+        return len(cls.visible_challenges()) > 0
+
     @property
     def path(self):
         return self.dojo.path / self.id
--- a/dojo_plugin/models/__init__.py
+++ b/dojo_plugin/models/__init__.py
@@ -405,14 +405,6 @@ class DojoModules(db.Model):
             resource.resource_index = resource_index
         self._resources = value
 
-    @hybrid_method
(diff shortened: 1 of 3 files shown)


dojo_plugin/models/__init__.py
```diff
@@ -405,6 +405,14 @@ def resources(self, value):
             resource.resource_index = resource_index
         self._resources = value
 
+    @hybrid_method
+    def visible(self):
+        return len(self.visible_challenges()) > 0
+
+    @visible.expression
+    def visible(cls, when=None):
+        return len(cls.visible_challenges()) > 0
```
ConnorNelson, line 414: Are you sure that this is valid sqlalchemy? At a glance it doesn't look like it would be since it iterates.

fpotier: I'm not, tbh I'm not familiar with sqlachemy. I just copied the `visible()` method of `DojoChallenges`

ConnorNelson:
Thanks for the testcase. I think this is the correct patch though. A module is still considered visible even if it doesn't have any visible challenges I believe (for example to see lecture materials, etc).

fpotier:
That makes sense, I didn't think about this case. Thank you

fpotier:
@ConnorNelson is there any case where a module should not/no longer be visible that would require another test case?

ConnorNelson:
I don't think so. Maybe eventually we'll want to support something like that, but for now this looks great. Thanks for the PR!

Issue #530, opened 2024-08-14
`local-tester.sh -D` fails


fpotier:
Using the `-D` option on `local-test.sh` fails:
```
docker run --rm --privileged -d -v /home/fpotier/dojo:/opt/pwn.college -v /home/fpotier/dojo/data:/data:shared -v /tmp/tmp.Io2JxJNnDM:/data:shared -p 2222:22 -p 80:80 -p 443:443 --name dojo-test dojo
docker: Error response from daemon: Duplicate mount point: /data.
```

PR #531, opened 2024-08-14
Fix for #530

fpotier:
I hardcoded the index but it might be better to iterate over the array and find the entry to remove.


zardus:
Thanks!

PR #536, opened 2024-08-18
Minor UI Bug Fix: double click when adding student id for first time 

etrickel:
Minor bug where the button does not submit on the first attempt when setting student id on the identity page.

This occurs because the user is not a student yet and the page dies because it tries to access official, which is only returned when the user is a student in the Dojo.

### DIFF
--- a/dojo_plugin/pages/course.py
+++ b/dojo_plugin/pages/course.py
@@ -296,7 +296,8 @@ def update_identity(dojo):
     db.session.commit()
 
     student = DojoStudents.query.filter_by(dojo=dojo, user=user).first()
-    if not student.official:
+    # FIX: not returning official after initially added on new user, it's not seeing user as student and thus it doesn't have the property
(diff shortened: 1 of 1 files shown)


ConnorNelson:
It's not a bug to become a student without having an ID in the roster. This supports the case for classes that don't have an explicit official roster, but want to be able to still track people with some identifier. It also makes it easier with students adding/dropping the class and not needing to immediately update the course roster so that they can identify themselves for later grading (imagine a student joins the class, but has to wait for you to update the roster). We still send a warning though (in case they typo their ID for example). I think this strategy minimizes the number of student emails (is there a better objective function? 🙂). Good catch on the Member -> Student bug!

etrickel:
What's the difference between a member and student then?
I guess, based on the code, I thought students were those that had the ID
field completed and authenticated whereas members were those that joined
the dojo. Is that not the case?

-Erik

PR #540, opened 2024-08-20
Increase width of UI on wider screens

etrickel:
Add three additional screen widths that will allow the UI to increase its width automatically when being viewed on higher resolution screens.


adamdoupe:
Why do the right margins in the first example look off? Is that just an artifact of the screenshot?

etrickel:
> Why do the right margins in the first example look off? Is that just an artifact of the screenshot?

I am not sure I see what you are getting at, but I re-did the screen shot comparison and drew a straight line down each side.

PR #545, opened 2024-08-28
Support Module level visibility

robwaz:
Adds support for module level visibility configuration.
- Module visibility cascades down to challenges 
- Hidden challenges are not included in challenge count for users
- Dojo Admin can see hidden modules and identify them as hidden
- Dojo Admin dojo challenge count includes hidden challenges

### DIFF
--- a/dojo_plugin/utils/__init__.py
+++ b/dojo_plugin/utils/__init__.py
@@ -167,23 +167,6 @@ def resolved_tar(dir, *, root_dir, filter=None):
     return tar_buffer
 
 
-def module_visible(dojo, module, user):
-    return (
-        "visibility" not in module or
-        module["visibility"]["start"] <= datetime.datetime.now(pytz.utc) or
-        is_dojo_admin(user, dojo)
-    )
-
-
-def module_challenges_visible(dojo, module, user):
-     return (
-         ("time_assigned" not in module and "visibility" not in module) or
-         ("time_assigned" in module and module["time_assigned"] <= datetime.datetime.now(pytz.utc)) or
-         ("visibility" in module and module["visibility"] <= datetime.datetime.now(pytz.utc)) or
-         is_dojo_admin(user, dojo)
-     )
-
-
 def is_dojo_admin(user, dojo):
     return user and dojo and dojo.is_admin(user)
 
--- a/dojo_plugin/utils/__init__.py
+++ b/dojo_plugin/utils/__init__.py
@@ -169,18 +169,19 @@ def resolved_tar(dir, *, root_dir, filter=None):
 
 def module_visible(dojo, module, user):
     return (
-        "time_visible" not in module or
-        module["time_visible"] <= datetime.datetime.now(pytz.utc) or
+        "visibility" not in module or
+        module["visibility"]["start"] <= datetime.datetime.now(pytz.utc) or
         is_dojo_admin(user, dojo)
     )
 
 
 def module_challenges_visible(dojo, module, user):
-    return (
-        "time_assigned" not in module or
-        module["time_assigned"] <= datetime.datetime.now(pytz.utc) or
-        is_dojo_admin(user, dojo)
-    )
+     return (
+         ("time_assigned" not in module and "visibility" not in module) or
+         ("time_assigned" in module and module["time_assigned"] <= datetime.datetime.now(pytz.utc)) or
+         ("visibility" in module and module["visibility"] <= datetime.datetime.now(pytz.utc)) or
+         is_dojo_admin(user, dojo)
+     )
 
 
 def is_dojo_admin(user, dojo):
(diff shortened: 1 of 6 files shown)


dojo_plugin/utils/__init__.py
```diff
@@ -169,18 +169,19 @@ def resolved_tar(dir, *, root_dir, filter=None):
...
-        "time_assigned" not in module or
-        module["time_assigned"] <= datetime.datetime.now(pytz.utc) or
-        is_dojo_admin(user, dojo)
-    )
+     return (
+         ("time_assigned" not in module and "visibility" not in module) or
+         ("time_assigned" in module and module["time_assigned"] <= datetime.datetime.now(pytz.utc)) or
+         ("visibility" in module and module["visibility"] <= datetime.datetime.now(pytz.utc)) or
+         is_dojo_admin(user, dojo)
+     )
```
ConnorNelson, line 184: Can we delete these two methods? They look like dead code to me. I think they are defined, imported, and then never used.

I know this is orthogonal, but it's better than refactoring dead code!

ConnorNelson:
Otherwise, it looks correct to me (I haven't manually tested).

This makes me think I was unnecessarily complex with these Visibility tables, and they should honestly just be columns, but I appreciate you matching the style! Probably (hopefully) some day this will be refactored into columns.

ConnorNelson:
Actually, edge case consideration!

What happens to the challenge when you define a module start, and a challenge start? Does the challenge take the `max()`, `min()`, or its own-specified value. If own-specified value, does this mean we can start a challenge that isn't visually available; should that be possible?

dojo_plugin/utils/__init__.py
robwaz: I don't believe we can delete these.

I had the exact same thought and initially did exactly that initially.  The `module_challenge_visible` is used in the course template to determine which modules will be included in grading.

The concept of a module's "time_assigned" must be distinct from the module visibility if we intend on supporting the use-case where a module is visible to students, but not yet "assigned" for grading purposes.  While I do not personally believe I will ever use it, someone could.  Additionally, removing it has the potential of breaking any dojos that currently rely the `time_assigned` field.  I could certainly rename one for a better read.

robwaz:
To the point of the "what if".  I haven't tested it, but applies visibility via the existing `visibility` functionality which takes a series checks in order: dojo, course, challenge and applies the widest visibility found.

This means it is possible to still support the use-case of a visible dojo with some challenges (or resources) hidden.  It is not possible to create a hidden dojo with challenges visible/accessible.  (Aside from challenges that are imported, but that's unrelated to this PR.)

https://github.com/pwncollege/dojo/blob/7ae88a6f9301bcf6ce205c2b9e598844e47e0f5c/dojo_plugin/utils/dojo.py#L257
https://github.com/pwncollege/dojo/blob/7ae88a6f9301bcf6ce205c2b9e598844e47e0f5c/dojo_plugin/utils/dojo.py#L293

robwaz:
The thing I do no know, is why this particular CI failure is happening.  It's failing to run practice mode in CI, but my local instance doesn't appear to have any issues and the error did not jump out at me reviewing the CI integration test logs.

dojo_plugin/utils/__init__.py
ConnorNelson:
```
$ git grep 'module_visible'
dojo_plugin/pages/course.py:from ..utils import module_visible, module_challenges_visible, is_dojo_admin
dojo_plugin/utils/__init__.py:def module_visible(dojo, module, user):
$ git grep 'module_challenges_visible'
dojo_plugin/pages/course.py:from ..utils import module_visible, module_challenges_visible, is_dojo_admin
dojo_plugin/utils/__init__.py:def module_challenges_visible(dojo, module, user):
```

Am I confused? Where is it? I think `time_assigned` is a dead concept.

robwaz:
> Otherwise, it looks correct to me (I haven't manually tested).
> 
> This makes me think I was unnecessarily complex with these Visibility tables, and they should honestly just be columns, but I appreciate you matching the style! Probably (hopefully) some day this will be refactored into columns.

I don't think that was the case.  This dynamic cascading property setup is more clear to me than if each concept (dojo, module, challenge) had their own column.  Suddenly the logic quite a bit messier with more edge-cases like challenges visible (API accessible I imagine) inside modules or dojos that aren't.  It's a lot of state to manage for no real win.

dojo_plugin/utils/__init__.py
robwaz: You are correct.

I'm going claim this is evidence that sleep does in fact have value.

ConnorNelson:
https://github.com/pwncollege/dojo/issues/546

I think this is likely not caused by your changes...

robwaz:
Let's find out if this works

Issue #544, opened 2024-08-22
Navbar flag input keys off of challenge id instead of challenge index


sjzhu:
## What?
The navbar flag input's previous and next buttons can sometimes show the wrong challenge if the current challenge was created recently, but shows up earlier in the challenge listing: 
<img width="514" alt="image" src="https://github.com/user-attachments/assets/20311f6b-c86f-413a-9c59-3c98a30b0f8a">

The line in question appears to be this one: https://github.com/pwncollege/dojo/blob/master/dojo_plugin/utils/dojo.py#L504
It probably should be using this: https://github.com/pwncollege/dojo/blob/master/dojo_plugin/models/__init__.py#L437

cc @ConnorNelson

JensHouses:
I'll Fix it. But this also means my tests are useless 🤣 sorry mate

sjzhu:
Haha! I was just working on this myself! I'll let the expert handle it though!

sjzhu:
Sorry, I lied! 😅  While learning how the dojo worked, I also saw a way I could reduce the number of queries sent to the DB, so I took a stab at it too!

PR #547, opened 2024-08-30
Refactor challenge ID comparison to use challenge index…

JensHouses:
…in `get_pre cur_next_dojo_challenge`

- Updated the query filters to compare `challenge_index` instead of `challenge_id`.
- Adjusted the order_by clauses to use `challenge_index` for sorting.

related to  Navbar flag input keys off of challenge id instead of challenge index #544


JensHouses:
grafik
no prob 4 minutes earlier :o

PR #548, opened 2024-08-31
added fastecdsa

prathamgupta36:
Adding fastecdsa which is a tool used for fast elliptic curve cryptography like digital signatures, it is being used to run the challenges in CTF Archive's cryptoctf2020, corctf2021, angstromctf2024, and corctf2022. 

PyPI link: [fastecdsa](https://pypi.org/project/fastecdsa/)

Github: [fastecdsa](https://github.com/AntonKueltz/fastecdsa)

### DIFF
--- a/challenge/Dockerfile
+++ b/challenge/Dockerfile
@@ -134,6 +134,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         upx-ucl
 EOF
 
+RUN pip install fastecdsa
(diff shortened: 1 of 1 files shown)


spencerpogo:
Instead of adding an additional layer to the image, could you add this to the `builder-tools-pip` step or to nix?

prathamgupta36:
sounds good, added in the `builder-tools-pip`

PR #556, opened 2024-09-03
[navbar] Fix challenge ordering and optimize queries

sjzhu:
## What?
Fixing this issue (https://github.com/pwncollege/dojo/issues/544) by reusing `get_current_dojo_challenge()` and the already created `challenge.module` and `module.challenges` (which helpfully sorts by challenge_index!)

This should also reduce the number of queries made to the DB, so hopefully we melt like 1% less or something.

## How was this tested?
Comparison of local dojo to prod:
Local:
image
Production:
image

cc @ConnorNelson


zardus:
Thanks!

PR #557, opened 2024-09-04
Made module completed percent easier to read at a glance. 

Smit2553:
Made the percentage of the module that has been completed always visible. I feel like it makes it easier to see at a glance. Still maintains that if 0% has been completed, the progress bar and text are not visible.

Local Changes:
image

Current Production:
image


Smit2553:
Highlighted modules are still more visible than if not hovered over. 

image

ConnorNelson:
Any opinions from @adamdoupe, the product manager of the progress bar.

PR #559, opened 2024-09-05
accept ecdsa-sha2-nistp256 public keys

livz:
The SSH server seems to support ecdsa-sha2-nistp256 keys, based on the output of `ssh -G`:

```
hostbasedacceptedalgorithms ssh-ed25519-cert-v01@openssh.com,...,ecdsa-sha2-nistp256,
..
kexalgorithms sntrup761x25519-sha512@openssh.com,..,ecdsa-sha2-nistp256
```

Modified the regular expression to allow ecdsa-sha2-nistp256 keys to be set in the UI.

This would be very useful because Secretive agent (which uses SSH keys stored in the TPM of Macs) only generates this key type - ecdsa-sha2-nistp256.

Hope this is the right place to make this change. 

Thank you!

### DIFF
--- a/dojo_plugin/api/v1/ssh_key.py
+++ b/dojo_plugin/api/v1/ssh_key.py
@@ -23,7 +23,7 @@ class UpdateKey(Resource):
         key_value = data.get("ssh_key", "")
 
         if key_value:
-            key_re = "ssh-(rsa|ed25519|dss) AAAA[0-9A-Za-z+/]{1,730}[=]{0,2}"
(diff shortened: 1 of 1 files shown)


ConnorNelson:
Solved by #715

livz:
Thanks @ConnorNelson ! I'm still getting the same error. Am I doing something wrong? (Although it looks like I'm getting the old message, before the change)
Screenshot 2025-07-16 at 09 01 03

ConnorNelson:
@livz try again, it should work now!

PR #575, opened 2024-09-24
Infrastructure: Migrate from NFS to ZFS for homes

ConnorNelson:


ConnorNelson:
Closed in favor of #607.

PR #583, opened 2024-10-03
Reword scoreboard to be clearer to some students

sjzhu:
## What?
At the request of @zardus , reword the scoreboard to be clearer that it is the past 7 days/30 days. Also change the heading based on what time frame we have selected.

## How was this tested?
Spinning up a local instance:
Screenshot 2024-10-02 191135
Screenshot 2024-10-02 191132
Screenshot 2024-10-02 191127

cc @ConnorNelson @robwaz @zardus


adamdoupe:
I like it, I'm not 100% sure about the dash between All and Time.

sjzhu:
Yeah, fair, I was waffling on it as well, it looked a little out of place without one, since the other two had it. I'll take a screenshot of it without the dash for comparison.

sjzhu:
image

zardus:
Awesome!

PR #584, opened 2024-10-03
Dynamic Course Configurable EC calculation

robwaz:
Note: This PR is coupled with the corresponding sensai changes which use the added API endpoints.

API changes:
- Add new API endpoint for bot to use to submit liked meme events
- Add new API endpoint for bot to use to submit thanks events
- Add new API endpoint for bot to get individual thanks count
- Add new API endpoint for bot go get individual memes count
- Add new API endpoint for obtaining thanks leaderboard

DB changes:
- Adds DiscordUserActivity, a generic table tracks actions users can apply to each other such as liking memes or thanking each other
- Adds several methods to DiscordUser to query DiscorduserActivity
- Adds `start_date` to Dojo object, which currently returns first earliest visible module as a proxy for course start

Course Grades page Changes
- Adds assignment type `helpfulness` and `memes`
  - Both of these have an optional, but necessary `max_credit` property 
    - sets individual max limits if course requires
    - Default is 1.00 max (100% of course)
  - `helpfulness` has a **functionally required** `method` property identifying the calculation method to use
    - Current valid values:
      - `1337log2` uses `1.337 ** log2(thanks) / 100 with a hard cap of `max_credit` if set
      -  `log50` uses log50 approaching the `max_credit` value
- Adds course.yml course level property `ec_limit`
  - Serves as a dynamic global cap on all extra credit (thanks, memes, and extra) 
  - Extra credit values on the grades page will be dynamically reduced to ensure total cannot exceed this threshold
- Adds dynamic progress indicators for helpfulness (total count) and memes (week ranges)
- Meme and helpfulness grade time frames are calculated as `dojo.start_date` (described above) to `start_date` + 16 weeks

I believe this setup is sufficient to represent all currently live ASU EC scenarios based on course syllabus.


dojo_plugin/api/v1/discord.py
```diff
@@ -56,5 +65,129 @@ def get(self, discord_id):
...
+
+    return get_user_activity_prop(discord_id, activity, start, end)
+
+def post_user_activity(discord_id, activity, request):
+    authorization = request.headers.get("Authorization")
+    res, code = auth_check(authorization)
+    if res:
+        return res, code
+
+    #data = request.get_json()
```
robwaz, line 115: @ConnorNelson 

Right here

robwaz:
Oh yeah.. and I also made DiscordUser IDs numbers.. because they are.

dojo_plugin/api/v1/discord.py
```diff
@@ -56,5 +65,129 @@ def get(self, discord_id):
...
+            if start_stamp is None:
+                start = year_stamp()
+            else:
+                start = datetime.datetime.fromisoformat(start_stamp)
+        except:
+            return {"success": False, "error": "invalid start format"}, 400
+
+        thanks_scores = DiscordUserActivity.query.with_entities(DiscordUserActivity.to_user_id, db.func.count(DiscordUserActivity.to_user_id)
+          ).filter(and_(DiscordUserActivity.timestamp >= start),
+                   DiscordUserActivity.activity_type == DiscordUserActivity.ActivityType.memes
```
sjzhu, line 174: @robwaz Should this be `DiscordUserActivity.ActivityType.thanks`?

sjzhu: Slight addendum: Are we opposed to a memes leaderboard? Doesn't need to be part of this, but just wanted to ask.

dojo_plugin/pages/course.py
```diff
@@ -109,9 +121,58 @@ def query(module_id):
...
+            if not discord_user:
+                return 0
+            thanks_count = discord_user.thanks_count(start=dojo.start_date())
+
+            return thanks_count
+
     def result(user_id):
         assessment_grades = []
 
+        ec_limit = dojo.course.get("ec_limit") or 1.00
```
sjzhu, line 145: Should the ec_limit changes be made to the course.yml's before this PR goes live? There might be a brief period where EC has a different cap than what's expected, some student might see it and then get confused.

sjzhu:
I have no authority to approve this, but LGTM other than that one apparent typo for the thanks leaderboard calc! 

> Couple things I realized this morning I need to address before merging.
> 
> * [ ]  DiscordAction Timestamp is "time of action", but we also want "time of target message" (ex: when meme was posted)
> * [ ]  Similarly, the meme grade calculation should use "time of message", not "time of action"
> * [ ]  Do we want thanks calculation to be "time of message" as well?  Less important, but that'd be consistent
> * [ ]  Update memes and thanks queries to be unique on `message_id` or (`channel_id`, `message_id`).  I do think it's worth storing duplicate actions in the DB for accurate data, but de-duping for grades.

Agree with the first 3 of these. Although I'm not sure what purpose the last point has? AFAICT, all discord message ids are unique, even between channels, so being unique on just `message_id` should be sufficient?

robwaz:
> Agree with the first 3 of these. Although I'm not sure what purpose the last point has? AFAICT, all discord message ids are unique, even between channels, so being unique on just `message_id` should be sufficient?

From the Discord API docs
```
These IDs are guaranteed to be unique across all of Discord, except in some unique scenarios in which child objects share their parent's ID.
```
You're probably correct that just being distinct on `message _id` is sufficient.  I was not aware the discord guaranteed uniqueness across all of discord (mostly).

dojo_plugin/api/v1/discord.py
robwaz: Good catch!  That's actually a fragment of a memes leaderboard I had.  I decided not to keep it because the memes leaderboard would need to not be limited to one meme per week. The backend already supports this.  However, from an impact side, I think encouraging quantity of memes over quality would lead to undesirable outcomes.

dojo_plugin/pages/course.py
robwaz: That `ec_limit` should only really be needed by fall cse365 currently.  All other courses have individual ec type limits that can be set with the `max_credit` property on the type.

robwaz:
@ConnorNelson any thoughts?  I'd like merge this in late tonight.

dojo_plugin/models/__init__.py
```diff
@@ -748,15 +753,78 @@ class SSHKeys(db.Model):
...
+    class ActivityType(enum.Enum):
+        thanks  = 1
+        memes   = 2
+
+    __tablename__ = "discord_user_activity"
+    id = db.Column(db.Integer, primary_key=True)
+    to_user_id = db.Column(db.BigInteger)
+    by_user_id = db.Column(db.BigInteger)
+    timestamp = db.Column(db.DateTime, default=datetime.datetime.utcnow)
+    activity_type = db.Column(db.Enum(ActivityType))
```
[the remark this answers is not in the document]
robwaz: You want the type to be a repeated string instead of an enum?

dojo_plugin/pages/course.py
```diff
@@ -24,6 +25,17 @@ def get_letter_grade(dojo, grade):
             return letter_grade
     return "?"
 
+def clamp_ec(limit):
+    global_limit = [limit]
```
ConnorNelson, line 29: Am I confused? Why is this a list?

dojo_plugin/models/__init__.py
```diff
@@ -264,6 +265,10 @@ def awards(self):
     def completed(self, user):
         return self.solves(user=user, ignore_visibility=True, ignore_admins=False).count() == len(self.challenges)
 
+    def start_date(self):
+        module_starts = [m.visibility.start for m in self._modules if m.visibility and m.visibility.start]
+        return min(module_starts) if module_starts else datetime.datetime(2024, 8, 22, 0)
```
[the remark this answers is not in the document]
robwaz: This magic number number is the start of the 2024 ASU fall semester, which is the first semester that would rely on this functionality.  Hence the default value.  We certainly can default it to the beginning of the year, but that would be a useless default.

dojo_plugin/pages/course.py
robwaz: This is some insane python thing with the decorator instance holding the limiting value across functions.  If it is not a list the value isn't captured in the decorator.

dojo_plugin/utils/discord.py
```diff
@@ -77,15 +77,19 @@ def get_discord_member(user_id):
     discord_user = DiscordUsers.query.filter_by(user_id=user_id).first()
     if not discord_user:
         return None
+    return get_discord_member_name(discord_user.discord_id)
```
ConnorNelson, line 80: Does `get_discord_member_name` even exist?

dojo_plugin/api/v1/discord.py
```diff
@@ -56,5 +65,129 @@ def get(self, discord_id):
...
+
+def post_user_activity(discord_id, activity, request):
+    authorization = request.headers.get("Authorization")
+    res, code = auth_check(authorization)
+    if res:
+        return res, code
+
+    #data = request.get_json()
+    try:
+        data = json.loads(request.get_data())
```
ConnorNelson, line 117: This has to be fixed. This is now how you bring json data in.

@bypass_csrf_protection
https://github.com/CTFd/CTFd/blob/c4b83305624a0f5f39700bb54b1e8711c90446f1/CTFd/plugins/__init__.py#L158
https://github.com/CTFd/CTFd/blob/master/CTFd/utils/initialization/__init__.py#L315

Should look just like this:
https://github.com/pwncollege/dojo/blob/5b4ee1db66557f3321da8122e8db9912d3a06cb2/dojo_plugin/pages/dojo.py#L135-L138

Please let me know if you are unable to get this to work.

robwaz: This does not work.

I believe it has to do with discord.py relying on `flask_restx` instead of flask.

ConnorNelson: https://github.com/pwncollege/dojo/commit/0dc67d2732055d1bbae650783455286b83dd8937

This commit should have resolved your issue. Don't worry about specifying `@bypass_csrf_protection`, CTFd automatically bypasses that logic for requests with an `Authorization` header anyways.

dojo_plugin/pages/course.py
ConnorNelson: This is how scoping works inside python, what you're looking for is the `nonlocal` keyword.

I believe this is what you are trying to express:

```py
def clamp_ec(limit):
    def decorator(func):
        def wrapper(*args, **kwargs):
            nonlocal limit
            result = func(*args, **kwargs)
            clamped_result = min(result, limit)
            limit -= clamped_result
            return clamped_result
        return wrapper
    return decorator
```

dojo_plugin/api/v1/discord.py
robwaz: And yet.. it still doesn't work.

```
  headers = {"Authorization": f"Bearer {DISCORD_CLIENT_SECRET}"}
        data = {"source_user_id": payload.user_id,
                "message_id": payload.message_id,
                "channel_id": payload.channel_id,
                "guild_id": payload.guild_id,
                "message_timestamp": message.created_at.isoformat(),
                }

        async with aiohttp.ClientSession() as session:
            response = await session.post(api_url, headers=headers, json=data)
            data = await response.json()
            print(data)
```

```
<@281565385139159040> liked <@281565385139159040>'s meme
{'message': "The server could not verify that you are authorized to access the URL requested. You either supplied the wrong credentials (e.g. a bad password), or your browser doesn't understand how to supply the credentials required."}
```

robwaz: Nvm.. I see that commit is in master...

robwaz:
YOLO

PR #589, opened 2024-10-08
Add static compilation support

### LINKED ISSUE #563: gcc cannot compile static binaries
```
hacker@practice~reverse-engineering~level1-0:/tmp$ cat do.c
#include <stdio.h>


int main() {
        char *name = "meowmeow";
        printf("hello world! %s\n", name);

        return 0;
}
hacker@practice~reverse-engineering~level1-0:/tmp$ gcc -static do.c
/nix/store/adakqsdbifx7d688z4874ap17clarhfn-binutils-2.41/bin/ld: cannot find -lc: No such file or directory
collect2: error: ld returned 1 exit status
hacker@practice~reverse-engineering~level1-0:/tmp$
```

Expected behavior
```
robwaz@fat-boy:/tmp$ cat do.c
#include <stdio.h>

int main() {
  char *name = "meowmeow";

  printf("hello %s!", name);
  return 0;
}
robwaz@fat-boy:/tmp$ gcc -static do.c
robwaz@fat-boy:/tmp$ file a.out
a.out: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=d0fff5f284ade28e16c734982d928d04b7a05201, for GNU/Linux 3.2.0, not stripped
robwaz@fat-boy:/tmp$
```

supercoolspy:
Fixes #563
Source: https://unix.stackexchange.com/questions/615679/how-to-compile-a-program-statically-on-nixos

### DIFF
--- a/workspace/additional/additional.nix
+++ b/workspace/additional/additional.nix
@@ -37,8 +37,6 @@ in
   packages = with pkgs; [
     (lib.hiPrio pythonEnv)
 
-    # Adds static compilation support
-    glibc.static
     gcc
     gnumake
 
--- a/workspace/additional/additional.nix
+++ b/workspace/additional/additional.nix
@@ -37,6 +37,8 @@ in
   packages = with pkgs; [
     (lib.hiPrio pythonEnv)
 
+    # Adds static compilation support
+    glibc.static
     gcc
     gnumake
 
--- a/workspace/flake.nix
+++ b/workspace/flake.nix
@@ -48,6 +48,8 @@
             curl
             findutils
             glibc
+            # Adds static compilation support
+            glibc.static
             glibcLocales
             gawk
             gnugrep
--- a/workspace/flake.nix
+++ b/workspace/flake.nix
@@ -48,7 +48,6 @@
             curl
             findutils
             glibc
-            # Adds static compilation support
             glibc.static
             glibcLocales
             gawk


supercoolspy:
Confirmed that it works on nixos

ConnorNelson:
```
bash-5.2# nix-shell -p gcc glibc.static glibc

[nix-shell:/]# echo -e '#include <stdio.h>\nint main() { puts("hello world"); }' > test.c

[nix-shell:/]# gcc test.c

[nix-shell:/]# ./a.out
Segmentation fault (core dumped)

[nix-shell:/]# gcc -static test.c

[nix-shell:/]# ./a.out
hello world
```

It seems like this causes non-static builds to not work?

ConnorNelson:
```
bash-5.2# nix-shell -p gcc glibc glibc.static

[nix-shell:/]# gcc -static test.c

[nix-shell:/]# ./a.out
hello world

[nix-shell:/]# gcc test.c

[nix-shell:/]# ./a.out
hello world
```

Hmmm interestingly, swapping the order fixes it. Saw someone talking about that [here](https://github.com/NixOS/nixpkgs/issues/59267#issuecomment-1435246886).


Can you move static to core next to the current glibc and make sure we get the order right?

ConnorNelson:
Thanks!

PR #592, opened 2024-10-14
Add check for nosuid data mount

spencerpogo:
Don't want this to happen to anyone else lol


zardus:
This doesn't add any requirements on `/data`, right? We already need that to be a mount point?

spencerpogo:
Yes, exactly. The `findmnt -n` check was already there. It will always be a mount point because docker will have to create one there.

PR #598, opened 2024-10-16
Infrastructure: add support for decomp2dbg in ghidra

private-static-voidmain:


zardus:
As @robwaz mentioned on discord, we are now in the Nix Era. Ghidra actually exists here: https://github.com/pwncollege/dojo/blob/master/workspace/additional/additional.nix

I'm not sure how to add decomp2gdb to nix, but @ConnorNelson, @supercoolspy, or @spencerpogo might!

ConnorNelson:
Looks like nix already has some support in place for adding extensions to ghidra: https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/ghidra/with-extensions.nix#L18.

supercoolspy:
It should be somewhat similar to adding this: https://github.com/NixOS/nixpkgs/blob/67f803dcf82ebe958a53fcfa7c1edb276558c3f6/pkgs/tools/security/ghidra/extensions/sleighdevtools/default.nix

supercoolspy:
Yeah, if you can test if it works and add the hash that looks good to me

ConnorNelson:
As we discussed, ida was broken by a missing icon file on their main site. This has been resolved in #611.

PR #600, opened 2024-10-18
Cap capstone version to 5.0.2 to avoid any kind of linking errors for freshly deployed pwncol instances 

bravosixgoingdark:
image

capstone v6 alpha had likely messed up capstone linking for many dojos. This issue often occurs when the instance is built from scratch


ConnorNelson:
Thanks!

bravosixgoingdark:
Yeah, should have been 5.0.3 but my limbo brain decided to type 2 into it

PR #605, opened 2024-10-22
Fixed challenge name overflow when using a small viewport

ThatCodingKMC:
Currently a lot of stuff displays slightly funny if the site is viewed on mobile, as a part of my efforts to fix that, this first contribution from me fixes text overflow in challenge names.

If you would prefer this change not cause the challenge name text to wrap around and instead just cut off with a "...", remove the "white-space: normal;" line from .accordion-item-name class.


zardus:
Thank you!

PR #608, opened 2024-10-25
Hide text in dojo description

etrickel:
The html filter for model and challenge descriptions now allows the "data-hide" attribute.
When <p "data-hide"="true">the text will not be visible but the text inside will be included in a copy of the text.

### DIFF
--- a/dojo_plugin/utils/__init__.py
+++ b/dojo_plugin/utils/__init__.py
@@ -143,6 +143,7 @@ def render_markdown(s):
         "b", "i", "strong", "em", "tt",
         "p", "br",
         "span", "div", "blockquote", "code", "pre", "hr",
+        "hide",
         "ul", "ol", "li", "dd", "dt",
         "img",
         "a",
@@ -152,6 +153,7 @@ def render_markdown(s):
         "*": ["id"],
         "img": ["src", "alt", "title"],
         "a": ["href", "alt", "title"],
+        "p": ["data-hide"]
     }
     clean_html = bleach.clean(raw_html, tags=markdown_tags, attributes=markdown_attrs)
     return Markup(clean_html)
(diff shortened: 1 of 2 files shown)


zardus:
Does our HTML filter filter out HTML comments? I've been using comments for this in the welcome dojo (e.g.: https://raw.githubusercontent.com/pwncollege/welcome-dojo/refs/heads/main/welcome/sensai/DESCRIPTION.md)

dojo_plugin/utils/__init__.py
```diff
@@ -143,6 +143,7 @@ def render_markdown(s):
         "b", "i", "strong", "em", "tt",
         "p", "br",
         "span", "div", "blockquote", "code", "pre", "hr",
+        "hide",
```
ConnorNelson, line 146: Are we sure this line does anything? I believe this would allow "<hide>" elements, which I don't believe you want (or exist).

etrickel:
It does filter them out, in my lab.
For example, I have
<h3>Objective</h3>
          <p>Create a program that takes words as command-line arguments,
converts each lowercase letter to its ROT13 equivalent, and prints the
transformed words to the console.</p>
          <p>ROT13 (rotate by 13 places) is a simple letter substitution
cipher that replaces a letter with the letter 13 letters after it in the
alphabet. Since the English alphabet has 26 letters, applying ROT13 to a
letter twice will return it to its original position. </p>
          
          

But that does not show up in the source file on pwn.college when I go to
group lab 4.

Has SENSAI escaped its container?

-Erik

dojo_plugin/utils/__init__.py
etrickel: it shouldn't have the "hide" tag
I went with using an attribute on p

etrickel: Ok, I removed, the hide tag, sorry about that.

PR #610, opened 2024-10-26
adding assessment.countdown_date to show students an alternative target date to the due date

etrickel:
This is a slight change from what I discussed with Connor, but I think it's more flexible because I do like the countdown timer.

With the integration with Canvas, I have a due date in Canvas and a pwn.college due date that comes two days later. Building in a slight grace period, with the intent, that they will treat the Canvas deadline as the target deadline. This gives them a bit of a grace period if something comes up; however, pwn.college counts down to the secondary date deadline and as the semester has gone on some students have started focusing on the pwn.college date and the countdown instead of the Canvas due date. 

 To prevent those students from doing that in the future, I have added a countdown_date functionality that will show the countdown date instead of the assessment date.  I plan to have the assessment's countdown_date match the Canvas date and make the pwn.college penalty date an unknown and random date that occurs after the Canvas date that way they'll focus on completing assignments by the due date listed in Canvas and have a bit of a grace period when things come up. 

-Erik

### DIFF
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -267,7 +267,8 @@ def view_module(dojo, module):
     if student or dojo.is_admin(user):
         now = datetime.datetime.now(datetime.timezone.utc)
         for assessment in module.assessments:
-            date = datetime.datetime.fromisoformat(assessment["date"])
+            assessment_date = assessment.get("countdown_date") if "countdown_date" in assessment else assessment["date"]
+            date = datetime.datetime.fromisoformat(assessment_date)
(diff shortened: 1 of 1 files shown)


zardus:
This might be easier to implement using a `grace_period` field than a `countdown_date` field, and just set a custom `grace_period` on each assignment. It's hidden to the users, only has to be implemented on the grades page, and makes more sense terms-wise

etrickel:
I'm not sure about easier since it's already written and was 2 lines.

Adding a grace_period wouldn't be too bad either though, it'd follow the
same logic as the extension field. If you think it makes more sense and
will get it approved faster then, I'm happy to change it.

-Erik

PR #612, opened 2024-10-27
Streamline startup and document architecture

zardus:
After catching up with the latest homefs changes, I started an ARCHITECTURE.md document to help people rapidly understand the DOJO. While writing it, I realized that the way we start the dojo is very convoluted (docker calls `dojo init` which calls `dojo-init` and then starts systemd, which in turn calls `dojo up`. I changed the startup flow to:

- Outer docker entrypoint is systemd
- Systemd has a `dojo-init.service` that calls `dojo-init`.
- After this, systemd has a `dojo-up.service` that calls `dojo-up`.

This makes things more sane, and also turns us into a two-stop-shop for the most important logs (`journalctl` for `-u dojo-init` and `-u dojo-up` and `docker logs ctfd` for ctfd logs).


zardus:
Failing with:

```
ctfd               | docker.errors.APIError: 400 Client Error for http+docker://localhost/v1.47/containers/0503714327bde0a6d7d8658cca4a539f4faceca6f16ac8a94711d3973eeed3f2/start: Bad Request ("failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting cgroup config for procHooks process: openat2 /sys/fs/cgroup/system.slice/docker-0503714327bde0a6d7d8658cca4a539f4faceca6f16ac8a94711d3973eeed3f2.scope/pids.max: no such file or directory: unknown")```

No idea why yet...

**Comment by zardus** [comment]
This is fucked.

File: dojo/dojo [hunk 1]
```diff
@@ -124,11 +124,12 @@ case "$ACTION" in
 
     # HELP: logs: display dojo logs
     "logs")
-        journalctl -u pwn.college -f --no-tail
+        journalctl -u pwn.college.* -f --no-tail
```
spencerpogo, line 127: should this be escaped?

zardus: Good catch!

PR #614, opened 2024-10-27
add a vagrantfile to give a foolproof example host for the dojo

zardus:
With our recent uptick in support requests, I think it'd be good to have a reference solution that cannot be messed up. Our outer docker is close, but host setup still gets in the way (e.g., br_netfilter). This might be a good thing to point to as the one setup we'll support.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 data/
+.vagrant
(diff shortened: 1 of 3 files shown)


bravosixgoingdark:
can this also works with other providers (virtualbox, VMware) or just only libvirt?

zardus:
I guess it'd only work with libvirt as-is. We'd need other provider entries for virtualbox and vmware.

PR #620, opened 2024-11-11
Workspace: Add support for challenge-provided code extensions

etrickel:
This adds the capability for a challenge to add it's own VSCode extensions.

### DIFF
--- a/workspace/services/code.nix
+++ b/workspace/services/code.nix
@@ -44,13 +44,21 @@ let
 
     until [ -f /run/dojo/var/ready ]; do sleep 0.1; done
 
+    
+    if [ -d /run/challenge/share/code/extensions ]; then 
+      EXT_DIR="/run/challenge/share/code/extensions"
+    else
+      EXT_DIR="@out@/share/code-service/extensions"
+    fi 
+
     ${service}/bin/dojo-service start code-service/code-server \
       ${code-server}/bin/code-server \
         --auth=none \
         --bind-addr=0.0.0.0:8080 \
         --trusted-origins='*' \
         --disable-telemetry \
-        --extensions-dir=@out@/share/code-service/extensions
+        --extensions-dir=$EXT_DIR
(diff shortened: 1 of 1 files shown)


spencerpogo:
Couple considerations. 

Could this have security implications because it lets challenges run arbitrary JS in the user's browser? I'm not really aware of how extension sandboxing works on the web version of VS Code. 

This would also override any extensions we install by default. Right now that's only `cpptools` so not too big of an impact, but that could grow in the future. My first thought would be to symlink from the challenge extension dir back to every default extension. Those symlinks would have to point to non-stable nix store paths, though, so that solution doesn't really work. The only alternative would be for the challenge to install every default extension as well, and keep that list in sync with the dojo.

etrickel:
Hi, 

You raise some good points. Thanks.

Fortunately, VSCode will still be housed inside the browser's sandbox. Thus, the top end of the available authority would be the same as any random web page. Nevertheless, I think it's more restrictive than that because on my local pwn.college instance, my test extension is unable to capture clipboard events like it can in an extension on VSCode's desktop version. I'm hopeful it's a restriction based on the fact that my local dev version is not using ssl. 

I will run a few tests this morning to see if an extension can fetch or otherwise access the DOM's API or the browser's global variables. 

Until May or so we didn't have any standard extensions outside of the builtin ones, right now there's two, the MS python and the MS cpptools extensions. I added to a .init to test this feature and as a part of that code it automatically creates a symlink to the current extensions by performing a quick search for them in the `nix` mess of files. If we needed to speed this up, we could leak it in the dojo-code script that's created. 

As the supported types of challenges expands, I imagine there'll be challenges that don't want any extensions or want completely different extensions. For example, as long as my class is still teaching scheme and prolog, it might be nice to have an extension installed to support those languages, but the other 99% of dojo's wouldn't want those extensions and their installation would be a waste of space.

PR #624, opened 2024-11-19
feat: Display detailed challenge completion info in hacker page

CeS-3:
**This PR adds functionality to display detailed challenge completion information in the user interface.**

**Purpose:**  
This update introduces the ability to show detailed information about the specific challenges a user has completed within each module of a dojo. It enhances the user experience by providing a clear and organized view of completed challenges along with their completion dates.

**Key Changes:**

- **Backend Updates:**  
  - Modified the `view_hacker` function to include detailed challenge completion data for each module and dojo.
  - Added the `user_solves` data structure to include completion dates for challenges in addition to their IDs.

- **Frontend Updates:**  
  - Updated the user interface template to display the names and completion dates of challenges under each completed module.
  - Ensured proper alignment and formatting of challenge rows for better readability.

### DIFF
--- a/dojo_plugin/pages/users.py
+++ b/dojo_plugin/pages/users.py
@@ -26,12 +26,25 @@ def view_hacker(user):
              .viewable(user=get_current_user())
              .filter(Dojos.data["type"] != "hidden", Dojos.data["type"] != "course")
              .all())
+    user_solves = {}
+    for dojo in dojos:
+        dojo_id = dojo.id
+        user_solves[dojo_id] = {}
 
+        for module in dojo.modules:
+            module_id = module.id
+            solves = module.solves(user=user, ignore_visibility=True, ignore_admins=False) if user else []
+
+            if solves:
+                user_solves[dojo_id][module_id] = {
+                    solve.challenge_id: solve.date.strftime("%Y-%m-%d %H:%M:%S") for solve in solves
+                }
     return render_template(
         "hacker.html",
         dojos=dojos, user=user,
         dojo_scores=dojo_scores(), module_scores=module_scores(),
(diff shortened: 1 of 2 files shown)


zardus:
We'll need to smoke test this with a production DB backup before merging. One of the reasons that the individual challenge details went away was that they were timing out the profile page. But that might actually have been because we tried to get this in a single query before, and this does a query per module, which might be more tractable? We'll have to see.

If we run into performance issues, I think the solution would be to add a global per-challenge score cache to https://github.com/pwncollege/dojo/blob/master/dojo_plugin/utils/scores.py, keep it warm in https://github.com/pwncollege/dojo/blob/master/dojo_plugin/scripts/warm_cache.py, and use that, but we can test this first before resorting to such measures.

CeS-3:
> We'll need to smoke test this with a production DB backup before merging. One of the reasons that the individual challenge details went away was that they were timing out the profile page. But that might actually have been because we tried to get this in a single query before, and this does a query per module, which might be more tractable? We'll have to see.
> 
> If we run into performance issues, I think the solution would be to add a global per-challenge score cache to https://github.com/pwncollege/dojo/blob/master/dojo_plugin/utils/scores.py, keep it warm in https://github.com/pwncollege/dojo/blob/master/dojo_plugin/scripts/warm_cache.py, and use that, but we can test this first before resorting to such measures.

Thank you for the feedback! I look forward to the test results and am happy to proceed with the next steps if needed.

zardus:
Fuck it, we'll do it live!

zardus:
Seems zippy!

zardus:
Thanks!

Issue #586, opened 2024-10-07
Support Arbitrary Grading Schemes


ConnorNelson:
Instructors want to grade in different ways.

One route is to say "No, this is not a feature/goal of DOJO. We will export to Canvas/other LMS, good luck!". But then you need a LMS, and historically we haven't reached for tools like Canvas because how hard is it to implement grading and deadlines? Probably this is not the path forward.

The issue of "arbitrary grading schemes" is that it's roughly equivalent to "arbitrary code". We could play sandbox games, and such, but ultimately when a student clicks on their grade page, they'd probably like to see the answer in <500ms. Sandbox games sound slow / a decent amount of code/infrastructure to facilitate. What are we going to do, launch a container on every grades endpoint hit? Static analyze the grading code? No thanks.

The best sandbox is the client browser. Load up their browser with the raw data, run some code locally. This should be fast and performant.

We could be sane and say "write javascript". But let's be real, we want to write python.
https://pyodide.org/en/stable/console.html

Then maybe we also do nightly batch jobs for instructors (maybe in a lame container), or maybe we can get away with just shipping all the data to the instructor's browser as well (that would be sweet).

PR #625, opened 2024-12-13
Added grace_period, fixed a bug with extra late date and extensions, …

etrickel:
Added @zardus suggested grace_period and fixed a bug with extensions and extra late date where a submission was not getting counted when it fell in between the two dates after an extension was added. 

While working on those items, I also improved the due date messages on the module page. 1) If we are past the first due date but there's an extra late due date then it will show that date. 2) If both dates are passed, it will now display that the due dates have passed instead of just being blank. 3) If the creator has `show_penalty` set in the `course.yml` then it will show the current flag values (i.e., the penalty adjusted percentage).

### DIFF
--- a/dojo_plugin/pages/course.py
+++ b/dojo_plugin/pages/course.py
@@ -42,7 +42,6 @@ def assessment_name(dojo, assessment):
 def grade(dojo, users_query, *, ignore_pending=False):
     if isinstance(users_query, Users):
         users_query = Users.query.filter_by(id=users_query.id)
-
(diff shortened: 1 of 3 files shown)


ConnorNelson:
I'm going to merge this now.

Going forward, the path to changing how grading works will be https://github.com/pwncollege/dojo/issues/586.

Ultimately, we have multiple instructors trying to use different grading schemes. Continuously increasing complexity in the dojo surrounding different grading paradigms / edge-cases is no good, and should be pretty nicely addressed by something like https://github.com/pwncollege/dojo/issues/586.

PR #629, opened 2025-01-03
Remove files

ConnorNelson:


ConnorNelson:
This change is blocked by welcome-dojo and example-dojo not depending on challenge-legacy.

zardus:
Is anything stopping us from making them depend on -legacy?

ConnorNelson:
We want to make it not require challenge-legacy so that CI doesn't depend on challenge-legacy. Otherwise, it will be slow (pulling a big image), and we may not even have the storage to do it. It shouldn't be too difficult, I think it'll basically just be `ubuntu:24.04 + exec-suid` and `python:latest + exec-suid`.

zardus:
Ah, okay. So we need to:

- make `pwncollege/challenge-exec-suid`
- translate all the challenges to use it
- test...

ConnorNelson:
I'm pretty sure the only reason this PR is failing tests is because it can't find `sudo`, as discussed in #1892. I am going to merge this PR, and address that there.

PR #635, opened 2025-01-09
Workspace: Fix ssh-entrypoint

ConnorNelson:


workspace/core/ssh-entrypoint.nix
```diff
@@ -4,7 +4,15 @@ let
   sshEntryPoint = pkgs.writeScript "ssh-entrypoint" ''
     #!${pkgs.bashInteractive}/bin/bash
 
-    $SHELL --login
+    export $(grep -v '^#' /etc/environment | xargs)
```
tesuji, line 7: I think this might be clearer, unless you wanna export other envs
```suggestion
    source /etc/environment
    export PATH
```

ConnorNelson: Thanks for the review, but yeah I want to export anything that is in there.

Now that I think about it though, this entrypoint is called from above with a `docker exec`, which is going to automatically result in all the environment variables from how we started the container existing, which is correct.

Probably if a custom image wants to specify more "global" environment variables, they should do so with `ENV` directives in their image, rather than needing to modify this file.

In other words, I'll go ahead and remove this line!

PR #637, opened 2025-01-11
Infrastructure: Fix multi-node networking

ConnorNelson:


ConnorNelson:
Hopefully this passes our test-cases, otherwise we're going to need to figure out a more clever way to get this behavior working on both single-node and multi-node.

On production, we manually have a patch deployed that looks something like this.

We add the network to `workspace-builder` to make sure `workspace_net` is created on the workspace nodes.

If `nginx-proxy` is in the network, this actually prevents it from correctly routing the packets to the other node. Not immediately obvious why, but I'm sure it's for simple networking reasons.

ConnorNelson:
Well, looks like I'll have to think about networking...

ConnorNelson:
As it stands, this PR makes multi-node work (and single-node fail).

Issue #593, opened 2024-10-14
Visibility should not 404 for current student


AlexanderEpolite:
With visibility on:
![image](https://github.com/user-attachments/assets/6c3dced8-53a6-4b25-afb4-850d5d5ebb69)

Visibility off:
![image](https://github.com/user-attachments/assets/3b8f8d64-64ce-4f49-ad32-d36258751f9f)

This page should still be visible if the hacker being accessed is the same as the one which is logged in.  When I go to my page, I see a 404 page when visibility is set to `Hidden`, which should not happen.

AlexanderEpolite:
To add, pwn.college/hacker returns a 404 when visibility is set to `Hidden`, even though I am logged in.

ConnorNelson:
I agree, this sounds like an issue.

Correct behavior would be that the visibility check for viewing your own profile is bypassed.

Are you also saying that you can’t view other users pages when your own visibility is hidden? Am I getting that right?

AlexanderEpolite:
> Are you also saying that you can’t view other users pages when your own visibility is hidden? Am I getting that right?

I am saying that I cannot view my own page when my visibility is hidden.  Sorry for any confusion, I used your page as an example for this issue.

Additionally, you cannot see yourself on dojo leaderboards when your visibility is hidden.

adamdoupe:
We should for sure fix the viewing your own page bug.

However, hidden is hidden, I don’t think it makes sense to see yourself on
the dojo leaderboards.

- Adam Doupé

ConnorNelson:
I agree, we can definitely fix the profile.

Unfortunately leaderboard would be trickier because we cache leaderboard calculations for performance reasons, and this would require some changes to that. Possibly we could track the most recent solve so that we could quickly recompute tie-breakers for ordering you, but then you'd see a different ranking than everyone else, and honestly I'm not totally sure that idea really makes sense.

AlexanderEpolite:
> Unfortunately leaderboard would be trickier because we cache leaderboard calculations for performance reasons, and this would require some changes to that.

I am not familiar with how the platform is written, but maybe it could be possible to compute all scores and cache them, then filter out the hidden ones (stored in cache) upon request, checking if the requesting user is equal to the hidden user.

PR #638, opened 2025-01-16
Fix 404 when viewing your own hidden profile

2stinkysocks:
Fixes https://github.com/pwncollege/dojo/issues/593, allows users to still view their own profile when it's marked as hidden.


2stinkysocks:
the issue also mentions leaderboards, I just omitted that part since it seems there was never a final decision on what to do

zardus:
Thanks!!

PR #639, opened 2025-01-17
fixing the canvas sync which stopped working. It appeared that using …

etrickel:
This commit addresses a critical issue where the synchronization process between pwn.college and Canvas broke due to changes in how Canvas handles SIS user IDs in the grade data submission.

**Issue:**
Previously, our system submitted grades directly using `sis_user_id` fields as identifiers in the grade data. Recent updates to the Canvas API have caused these submissions to fail, possibly due to changes in how Canvas handles external identifiers.

**Resolution:**
To resolve this issue, the following changes were implemented:
- **User Retrieval Update:** Modified the grade synchronization script to first make a call to the Canvas API to fetch all users associated with the course. This is done using the `/api/v1/courses/{course_id}/users` endpoint, which returns all user data, including both `sis_user_id` and Canvas's internal `id`.
- **ID Mapping:** Created a mapping from `sis_user_id` to Canvas's internal user ID (`id`). This map is now used to translate the SIS IDs in our system to the corresponding Canvas IDs when submitting grade data.
- **Grade Submission Adjustment:** Updated the grade submission logic to use Canvas's internal user IDs instead of SIS IDs. This involved adjusting the payload of the API call to the `/api/v1/courses/{canvas_course_id}/assignments/{assignment_id}/submissions/update_grades` endpoint, ensuring that grades are correctly associated with user accounts in Canvas.

### DIFF
--- a/dojo_plugin/pages/canvas.py
+++ b/dojo_plugin/pages/canvas.py
@@ -15,7 +15,6 @@ canvas = Blueprint("canvas", __name__)
 
 logger = logging.getLogger(__name__)
 
-
 def canvas_request(endpoint, method="GET", *, dojo, **kwargs):
     missing = [attr for attr in ["canvas_token", "canvas_api_host", "canvas_course_id"] if not (dojo.course or {}).get(attr)]
     if missing:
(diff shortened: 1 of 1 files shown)


zardus:
@ConnorNelson is this mergeable?

PR #654, opened 2025-02-05
Infrastructure: Add devcontainer

ConnorNelson:


ConnorNelson:
For some reason, there is an issue where the devcontainer itself can reach the internet, but the subcontainers cannot.

https://github.com/docker-library/docker/issues/467 discusses this issue, which seems to be the same issue.

As far as I can tell, it only happens because codespaces is on an Ubuntu 22.04 host? This clearly doesn't happen on github actions. It might also be (somehow) related to the devcontainer running with a `--net=host`? I'm not confident that is true, but it seems to be the default behavior (based on receiving a 10. ip address instead of a "normal" docker 172.). Maybe it's both of those as the same time? Not super interested on diving even deeper into this.

PR #660, opened 2025-02-17
Add API endpoint to list available dojos

fpotier:
Adds a `/pwncollege_api/v1/dojo/list` endpoint that returns the list of available dojos.

To follow REST API best practices it might be better to use this path for the resource: `/pwncollege_api/v1/dojos`


zardus:
Thanks!

zardus:
I went ahead and moved it to `dojos`, per your suggestion.

zardus:
With some tweaks, this is now live on pwn.college! Thanks again!

PR #666, opened 2025-03-02
Add share buttons to hacker profile

alyssaditroia:
Adds social share buttons to the hacker profile page, visible only when the logged-in user is viewing their own profile.
Changes Made
Removed open graph comments on popup.js
Added social media share buttons to the current user's hacker profile under belts, badges, website, etc.
Buttons do not appear when viewing another user's profile.
How I tested
Created an additional user to verify that share buttons do not appear when viewing other users' profiles.
Confirmed placement below all other profile details.
Unable to test the impact on the dojo section but expect it to be properly contained within jumbotron.


ConnorNelson (changes requested):
Looks great, I think you're just missing `_external=True`.

Right now we just link to, for example, https://linkedin.com/share?url=/hacker/1 (which doesn't include the full URL like it needs to).

PR #675, opened 2025-03-18
Update index.html

hidehiroanto:
index.html doesn't use markdown


zardus:
I swear, I do this every time...

PR #677, opened 2025-03-20
Workspace: Remove suid interpreter

ConnorNelson:
The suid interpreters have been moved to challenge-legacy, where they belong.

https://github.com/pwncollege/challenge-images/commit/ba00d0d427b54aa31061aa3f2fe86f10ce2fe391

---

We still need sudo to be suid though, and so we use exec-suid for that now.


ConnorNelson:
We already have a `sudo` test that should correctly sanity check this PR.

ConnorNelson:
Unfortunately we have to make an exception in workspacefs to allow `sudo` to work, otherwise `exec-suid` fails to parse `sudo`'s shebang.

PR #678, opened 2025-03-20
Infrastructure: Fix codespaces docker networking

ConnorNelson:
For some reason, codespaces has an issue where networking does not work in the subcontainers.

This is quickly obvious when `workspace-builder` never completes, and so the infrastructure does not start.

It's not clear why. We have `nf_tables` correctly loaded, so I'd assume it would work, and yet it doesn't.

Switching to legacy iptables (and restarting docker service, if it's already running) resolves this.


ConnorNelson:
Using a simplifier version of the solution at https://github.com/docker-library/docker/pull/468#discussion_r1442131459.

PR #679, opened 2025-03-22
Workspace: Prevent burpsuite from trying to mount

ConnorNelson:


ConnorNelson:
This got merged, but GitHub decided to not mark this as merged...

I definitely blame ASU wifi.

PR #681, opened 2025-03-27
Site: Support progression locked challenges

maxtrap:
# Features added

Dojo creators will now be able to set an optional flag "progression_locked" under each challenge inside the module.yml. "progression_locked" takes a boolean as its value like so:

image

Challenges with "progression_locked: true" will appear as locked to the user if the user has not solved the immediately prior challenge. Once the user solves the prior challenge, it will unlock via javascript (no need to refresh the page).

# Notes

- Users will not be able to start progression locked challenges prematurely, even if they remove the disabled attribute inside the html, or even if they forge a start challenge request.
- Dojo admins will see progression locked challenges as unlocked, with small text indicating that the challenge is progression locked. 
- If the first challenge in the module is progression locked, it will ignore the lock completely. In essence, the first challenge of the module will always be unlocked.

# Limitations

- The locking functionality is done client-side to some extent, meaning the descriptions of locked challenge are still sent to the user. There is nothing preventing the user from removing the "disabled" class from the challenge button html element, enabling the user to read the challenge descriptions. However, the user is barred from actually starting the challenge.

### DIFF
--- a/dojo_plugin/utils/dojo.py
+++ b/dojo_plugin/utils/dojo.py
@@ -334,7 +334,7 @@ def dojo_from_spec(data, *, dojo_dir=None, dojo=None):
                     challenge=challenge(
                         module_data.get("id"), challenge_data.get("id"), transfer=challenge_data.get("transfer", None)
                     ) if "import" not in challenge_data else None,
-                    progression_locked=challenge_data.get('progression_locked'),
+                    progression_locked=challenge_data.get("progression_locked"),
                     visibility=visibility(DojoChallengeVisibilities, dojo_data, module_data, challenge_data),
                     default=(assert_import_one(DojoChallenges.from_id(*import_ids(["dojo", "module", "challenge"], dojo_data, module_data, challenge_data)),
                                         f"Import challenge `{'/'.join(import_ids(['dojo', 'module', 'challenge'], dojo_data, module_data, challenge_data))}` does not exist")
--- a/dojo_plugin/utils/dojo.py
+++ b/dojo_plugin/utils/dojo.py
@@ -334,7 +334,7 @@ def dojo_from_spec(data, *, dojo_dir=None, dojo=None):
                     challenge=challenge(
                         module_data.get("id"), challenge_data.get("id"), transfer=challenge_data.get("transfer", None)
                     ) if "import" not in challenge_data else None,
-                    progression_locked=True if challenge_data.get('progression_locked') else False,
+                    progression_locked=challenge_data.get('progression_locked'),
                     visibility=visibility(DojoChallengeVisibilities, dojo_data, module_data, challenge_data),
                     default=(assert_import_one(DojoChallenges.from_id(*import_ids(["dojo", "module", "challenge"], dojo_data, module_data, challenge_data)),
                                         f"Import challenge `{'/'.join(import_ids(['dojo', 'module', 'challenge'], dojo_data, module_data, challenge_data))}` does not exist")
--- a/dojo_plugin/utils/dojo.py
+++ b/dojo_plugin/utils/dojo.py
@@ -89,6 +89,7 @@ DOJO_SPEC = Schema({
             Optional("image"): IMAGE_REGEX,
             Optional("allow_privileged"): bool,
             Optional("importable"): bool,
+            Optional("progression_locked"): bool,
             Optional("auxiliary", default={}, ignore_extra_keys=True): dict,
             # Optional("path"): Regex(r"^[^\s\.\/][^\s\.]{,255}$"),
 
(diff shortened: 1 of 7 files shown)


zardus:
Let's also do a testcase for this!

dojo_plugin/utils/dojo.py
```diff
@@ -333,6 +334,7 @@ def import_ids(attrs, *datas):
                     challenge=challenge(
                         module_data.get("id"), challenge_data.get("id"), transfer=challenge_data.get("transfer", None)
                     ) if "import" not in challenge_data else None,
+                    progression_locked=True if challenge_data.get('progression_locked') else False,
```
ConnorNelson, line 337: I think this can be simplified to:
```py
progression_locked=challenge_data.get("progression_locked")
```

dojo_theme/templates/module.html
```diff
@@ -64,20 +64,34 @@ <h4 class="accordion-item-name">{{ resource.name }}</h4>
...
   <div class="accordion" id="challenges">
-    {% for challenge in challenges %}
+    {% for i in range(0, challenges|length) %}
+      {% set challenge = challenges[i] %}
+
       {% set solved = "challenge-solved" if challenge.challenge_id in user_solves else "challenge-unsolved" %}
       {% set active = "challenge-active" if challenge.challenge_id == current_dojo_challenge.challenge_id else "" %}
-      {% call(header) accordion_item("challenges", loop.index) %}
+      {% set previous_challenge = challenges[i - 1] %}
+      {% set locked = challenge.visible() and challenge.data['progression_locked'] and previous_challenge.challenge_id not in user_solves and solved == "challenge-unsolved" %}
```
ConnorNelson, line 73: You should be able to just do `challenge.progression_locked` (data_fields work like this).

ConnorNelson: Let's also just special case the first challenge to not be locked even if it's progression_locked. Probably `loop.first` will be useful here.

dojo_theme/templates/module.html
```diff
@@ -64,20 +64,34 @@ <h4 class="accordion-item-name">{{ resource.name }}</h4>
   <h2>Challenges</h2>
 
   <div class="accordion" id="challenges">
-    {% for challenge in challenges %}
+    {% for i in range(0, challenges|length) %}
+      {% set challenge = challenges[i] %}
```
ConnorNelson, line 68: Lets stick to the original jinja of `{% for challenge in challenges %}`, I understand you want to get at `i`, you do that with `loop.index`, or actually in this case `loop.index0` (which is 0-based vs the default 1-based).

See docs [here](https://jinja.palletsprojects.com/en/stable/templates/#:~:text=The%20current%20iteration%20of%20the%20loop,-.).

ConnorNelson (changes requested):

ConnorNelson:
Let's also add a check about [here](https://github.com/pwncollege/dojo/blob/041e6682761daddeb4bb0aaa0de8671589fd95e3/dojo_plugin/api/v1/docker.py#L359) to see if it should be possible to start the challenge.

Roughly speaking, you'd take the `dojo_challenge`, grab it's module and index, index to previous challenge, and confirm it is solved by the user. Roughly speaking you could test to see if it's solved with something like:
```python
Solves.query.filter_by(user=user, challenge=previous_dojo_challenge.challenge).first()
```
We would only want to make these queries if the challenge *is* progression_locked, and also make sure it isn't the first challenge in the module.

ConnorNelson:
We can probably just accept the limitation that the description is only hidden client side for now, but it would definitely be great to prevent challenge start.

ConnorNelson:
Notice also the `dojo.is_admin()` check a few lines up [here](https://github.com/pwncollege/dojo/blob/041e6682761daddeb4bb0aaa0de8671589fd95e3/dojo_plugin/api/v1/docker.py#L351). We'd want to do the same so that dojo admins can bypass.

Additionally, dojo admins should ideally see all the challenges as unlocked on the client side. Maybe we still have a little lock instead of a flag, but the description is accessible. Maybe borrowing ideas from [here](https://github.com/pwncollege/dojo/blob/041e6682761daddeb4bb0aaa0de8671589fd95e3/dojo_theme/templates/module.html#L77).

dojo_plugin/utils/dojo.py
maxtrap: Okay, my motivation for that was to avoid having null values in the database, but I this seems to work fine too.

dojo_theme/templates/module.html
maxtrap: That's good to know. I hadn't realize Jinja was its own thing I assumed it was just using python syntax.

maxtrap:
To your last point, I was brainstorming ways to display progression locked challenges to the administrator without conflicting with the "hidden - you are this dojo's administrator" text. My friend suggested to make the "you are this dojo's administrator" appear at the top and each challenge would be labelled with its modifications

This is how I designed it

image

Just wanted to know your thoughts or if you had a different way of doing it in mind.

zardus:
What about doing it the same way that we handle admins being able to see
hidden challenges currently?

maxtrap:
Would this work better? (Challenges are hidden at random for PoC)

Screenshot from 2025-03-30 16-22-27

My main issue is trying to make sure challenges that are both hidden and progression locked are displayed cleanly.

zardus:
Looks great to me!

PR #682, opened 2025-03-29
Modified scoreboard to display school logo for users with school emails like ASU users (added Big 12 logos)

JamesNKing:
School logos use the same logic as the ASU logo code. 

Didn't get the chance to setup a full local pwn.college instance yet so this code will need to be run in test environment to make sure it renders properly.


JamesNKing:
If this implementation ends up working, I can expand the school list to beyond the Big 12, and we will also probably need to move the email_symbol_asset() code to its own file since it'll turn into a mile long list of match case statements.

hidehiroanto:
Why not just do something like this?
```py
def email_symbol_asset(email):
    email_domain = email.split("@")[1]
    if email_domain.endswith(".edu"):
        school = email_domain[:-len(".edu")]
        schools = get_schools() # get list of schools somehow
        if school in schools:
            group = f"schools/{school}.png"
        else:
            group = "student.png"
    else:
        group = "hacker.png"
    return url_for("views.themes", path=f"img/dojo/{group}")
```
This would require renaming every image to the school’s domain, including fork.png to schools/asu.png

robwaz:
I agree with Hanto's suggestion as far as addressing an endless if/else chain.  `get_schools` could look for a similarly named file in the schools directory and return that.

On a separate note, I think it makes the scoreboard more exciting!  Here's a screenshot of the inital PR's appearance.
Screenshot 2025-03-29 at 01 22 45

We might need to either standardize the image size or resize it ourself.  My preference would be the former.  If we standardize on a set size png I think all the images will fit nicely.

JamesNKing:
I added the change to make the images have the school email domains as their name to make it easier to match them. As for the image size standardization, I don't do much front end but I threw in an attempt in the CSS.

ConnorNelson:
Hey, thanks for contributing! I think the idea of some group rivalry/pride/etc sounds fun.

A few thoughts:

- I'd have some privacy concerns. Would the average user expect that by giving an email of `me@random.edu` (which is private) means everyone gets to know I am affiliated with `random.edu`? We do this with ASU, but ASU is special because 1. pwn.college was born out of ASU and so 2. we have **many** ASU users, so probably this is a bit more blending into the crowd.

- I'd also have some potential trademark concerns. I know that universities can be very particular about their logos being used. I'm not totally sure what is or is not permissible. I have no doubt that we can reference lots of universities, but I'm much less sure that we can do it via their logo.

- In a world where we slowly just bring in every school's logo, at some point it feels like maybe the right solution is to just go all the way and allow arbitrary profile pictures? We could also conveniently default to Discord profile pictures. Alternatively, would we have to then story being concerned about some "weirder" profile pictures being prominently displayed in standard public locations? I might also be in favor of even just entirely eliminating that column; maybe the more interesting thing is just your belt/badges.

- Something to keep in mind is that you can currently see who has an `edu` email, and it's relatively rare. Specifically, with the exception of ASU, right now we have in total 76 users that have an email belonging to a Big 12 school.

I do think there is something compelling about this group rivalry/pride/etc sort of idea, I'm just not totally sure what the right way is to express it visually or otherwise.

I'd be super interested if you have more thoughts. Thanks!

JamesNKing:
You bring up some good points. These policy decisions are ultimately up to ASU and the pwn.college maintainers, but here are my thoughts: 

* In regard to privacy, my motivation for this idea stemmed from the fact that university students are already identified as such in the form of the graduation hat icon that is placed on accounts with .edu emails. Because students are aware of this, I think there is a certain degree of consent that is being provided when someone creates a pwn.college account using their school email. That being said, I'm sure adding a toggle in the user profile for whether or not a user _wants_ to have their school logo displayed wouldn't be out of the realm of possibility. 

* As for trademark concerns, my understanding is that this would fall under fair use. The pwn.college project is a free, educational, open source platform, and the use of university logos I believe would be categorized as "nominative fair use" as the logos would strictly be used to identify user accounts who have supplied university-affiliated email accounts. Additionally, the use of a broad range of university logos would help to dissuade any potential suggestion of specific endorsement by any individual university. A text disclaimer located somewhere on the site to clarify that the site isn't claiming endorsement by any of the universities whose logos appear may be appropriate. As with all legal matters, if concerns remain about potential liability then it would be best to consult legal expertise. 

* In regard to profile customization, I would actually argue against arbitrary user profile customization, specifically due to the educational association and nature of the pwn.college platform. University logos would be able to provide some nice visual diversity and distinguish university users on the platform (and provide for some school bragging rights I imagine), all while maintaining a more formal appearance consistent with academic content. Popular educational websites like Khan Academy also limit user profiles to a provided list of designs. That being said, if full user profile customization is something that wants to be added to the list of planned site upgrades, I don't see any issue with continuing to use university logos as a starting point implementation. 

* My motivation with starting with the Big 12 schools for logos was that both ASU and my school are in the Big 12, so I figured it would be a good data set to start with when building the initial design. The process for collecting the school domains and their official logos wasn't very hard, and I would certainly plan to collect that data for all of the major conferences (at least) since it only has to be done once and then the data is available any time.

JamesNKing:
Something else to consider: as always, the safest bet when it comes to copyright/trademark concerns is to just get permission from the owner. This approach would obviously be the most amount of work compared to any of the others, but could provide some additional peace of mind for anyone who might be worried about pwn.college's ability to use the logos. 

I imagine the best way to accomplish that would be to draft some sort of request letter outlining details of who is asking for permission and what the specific intended use would be. Then, slowly roll out each logo on the site as each school responds with approval (if they do provide approval). 

Besides the time this would likely take, another downside of this approach could be that by making a formal request, it may force the schools to formally respond. By this I mean they may be forced to reject the request or very narrowly provide approval based on their legal counsel's guidelines, even if said schools didn't otherwise have any issue with the logo use prior to be asked to formally decide. This scenario, of course, is only potentially relevant if the pwn.college maintainers decide they believe the use case for the logos _does_ in fact have a good argument for fair use. 

Lastly, whether or not a request letter is sent, it would likely be fairly safe (again, in my non-expert legal opinion) to use the logos to the best of our understanding of fair use law, including adding disclaimers, and simply being prepared to take any or all of the logos down fairly quickly should a school or other entity express disapproval of the use. Since pwn.college is free and educational in nature, any legal action would essentially be limited to requesting that the logos be removed, as it would be hard to argue that a free learning platform had some sort of monetary benefit from the use of 3rd party material.

robwaz:
We had a discussion about this feature.  The takeaway was that a few changes are needed in order to merge it in.

- [ ] The affiliation image must be consistently formatted/placed on the scoreboard.
- [ ] Allowed images are pre-approved static images hosted on the dojo itself.
- [ ] The affiliation image must be user selectable rather than be linked to the registered email.

This means it needs a bit of UI work to somehow allow the user to select the affiliation avatar image.  Similarly, we'd need to create an "affiliation image" database table as well as add a reference column to the users table to indicate the user's selected affiliation avatar preference.

PR #683, opened 2025-03-31
Site: Add basic search functionality

visargsoneji:
It allows users to search across names and descriptions of dojos, modules, and challenges.
Screenshot 2025-03-30 at 4 50 30 PM
Screenshot 2025-03-30 at 4 50 45 PM
Screenshot 2025-03-30 at 4 51 15 PM
Screenshot 2025-03-30 at 4 56 28 PM

### DIFF
--- a/dojo_plugin/api/__init__.py
+++ b/dojo_plugin/api/__init__.py
@@ -10,6 +10,7 @@ from .v1.scoreboard import scoreboard_namespace
 from .v1.ssh_key import ssh_key_namespace
 from .v1.workspace_tokens import workspace_tokens_namespace
 from .v1.workspace import workspace_namespace
+from .v1.search import search_namespace
 
 
 api = Blueprint("pwncollege_api", __name__)
@@ -24,3 +25,4 @@ api_v1.add_namespace(scoreboard_namespace, "/scoreboard")
 api_v1.add_namespace(ssh_key_namespace, "/ssh_key")
 api_v1.add_namespace(workspace_tokens_namespace, "/workspace_tokens")
 api_v1.add_namespace(workspace_namespace, "/workspace")
+api_v1.add_namespace(search_namespace, "/search")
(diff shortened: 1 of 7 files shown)


ConnorNelson:
Wow! Very cool. This looks like a great start.

A couple things before we merge this initial prototype:
1. Can we get a few testcases added? Look at some of our existing tests, in particular [`test_get_dojo_modules`](https://github.com/pwncollege/dojo/blob/041e6682761daddeb4bb0aaa0de8671589fd95e3/test/test_running.py#L74).  Let's get a test that confirms that all 6 of `{dojo, module, challenge} x {name, description}` correctly search; feel free to just make sure the API works.

2. How concerned are we about performance? We don't need to way-overthink this (at least not yet), we can adapt as we find out. But maybe a ~200ms debounce logic would make sense? In other words, don't hit the endpoint until they stop typing for 200ms. That should get us a quick major win with very-minimal UX loss.
```js
const debounce = (fn, delay) => {
  let timer;
  return (...args) => {
    clearTimeout(timer);
    timer = setTimeout(() => fn(...args), delay);
  };
};

const handleInput = event => {
  console.log("User typed:", event.target.value);
};

input.addEventListener("input", debounce(handleInput, 200));
```

3. Any thoughts on how we can display descriptions when those match? It would be interesting to see how other search systems handle this.

---

When you get 1 and 2 done, this will absolutely be ready to merge/deploy to production. I think 3 will be the next step in a follow on PR.

This is great, thanks!

visargsoneji:
Hi @ConnorNelson,

1 and 2 are both taken care of — we now have integration tests covering all 6 cases across dojo/module/challenge × name/description, and the frontend uses a 200ms debounce to reduce unnecessary load without impacting UX.

For 3, a possible approach is to include a matched_in field or a trimmed description_snippet in the API response when a match occurs in the description. We can then conditionally render that snippet below the result name (muted/small) only when it's relevant. Many search systems (like ChatGPT) follow this pattern to keep things clean but informative.

Issue #667, opened 2025-03-04
Data Collection on Flag Submission


ConnorNelson:
I want to collect various bits of data on flag submission.

For example, thumbs up thumbs down "I liked this challenge" after submitting flag.

This could be 100% of the time for all users, or randomly 10%, or whatever.

Could be likert questions 1-5 "strogly agree ... strongly disagree".

Could be free form questions.

Could be contents "show me you understand..." vs preference data.

2stinkysocks:
After many hours of messing around with this feature (user defined html surveys on a per-challenge basis, sandboxed to prevent xss), I think I have a solid list of reasons why this is probably not something we should implement:

- Style issues: users can obviously redefine any css they want inside their survey, which would place the responsibility of updating survey styles if any changes to the global style of the site are made onto the dojo owner and not the pwncollege maintainers. One way to make this at least a little less manual is to have a script that copies pwncollege's styles into the survey iframes on page load, which would work, but any external fonts or links that require an http request need to be made by a script before placing the styles into the iframe, since the requests would be cross-origin. On top of this, either we'd have to put something into place to enforce the look and feel of pwncollege, or be okay with users making surveys that don't match it at all.
- Coupling issues: A js script would have to be written in order to properly load assets into the iframe (trying to decouple/prevent excess code duplication as much as possible by not just duplicating the entire css of pwncollege into each iframe in the dojo definitions), which I don't think is possible to write in a way that wouldn't just break if changing the styles; it would require a lot more work if ever changing the styles/fonts/etc of the main site, and everything I tested myself was far from being robust enough for me to feel comfortable using.
- Security issues: The implementation I've been testing _seems_ like it's secure, but I'm concerned there will be a vulnerability that I miss. I've been testing a sandboxed iframe that communicates with a webworker to make only a specific survey-submit request to the backend. This is more of just a concern and not a real issue, since there shouldn't actually be any vulnerability with this as far as I can tell, but there always is a possibility.
- Motive issues: To my knowledge, we will be the only people who will be using this feature (in it's current state) anyways, since all data is only accessible to pwncollege admins (for now). If we really wanted to define a new survey type, it should be almost the exact same amount of effort to just make a quick pr to pwncollege to add the html (backend likely won't need to change). I don't really see any reason why for our specific use case this wouldn't be ideal other than having to re-deploy the dojo


[...]

PR #684, opened 2025-04-01
Research: Add surveys on challenge solve

2stinkysocks:
Adds api and frontend for optional surveys presented to users after they complete a challenge. 

Implements https://github.com/pwncollege/dojo/issues/667


2stinkysocks:
The feature works based on my own testing, for now the response data is just stored in a table `survey_responses` by user, module, challenge, dojo, and timestamp

PR #686, opened 2025-04-03
Engagement: Challenge Solves Activity Tracker on Hacker Profile

alyssaditroia:
* 365-day challenge solve activity graph. 
* Uses percentile-based color levels based on users max solves per day.
* Daily solve counts are computed using local timezone.
* Graph has been tested across multiple device sizes with DevTools.
* Repeated DOM queries are cached to improve performance.


ConnorNelson:
Great, thanks!

PR #692, opened 2025-04-09
Site: Enhance Search with Description Snippets and Keyboard/Mouse Navigation

visargsoneji:
- Search results now include matching description snippets to provide better context and build user trust.
- Users can navigate results using the ↑ / ↓ arrow keys, and press Enter to open the selected item.
- Mouse hover highlights results just like keyboard navigation, creating a seamless interaction between input methods.

Screenshot 2025-04-08 at 7 06 31 PM


ConnorNelson:
Thanks!

PR #713, opened 2025-06-01
Research: user-defined surveys

2stinkysocks:
Some initial changes for user-defined surveys. Currently css is NOT sanitized, this needs to be changed before merging. Also need to test whether stylesheets within the surveys will conflict with each other and do some more thorough testing.

This is implemented by using bleach to sanitize survey html before it's inserted into the db, then rendering it inside a form. I also implemented a custom attribute that can be used by surveys to instantly submit onclick with a specified value (ex: thumbs up/down in a single click).

### DIFF
--- a/dojo_plugin/utils/dojo.py
+++ b/dojo_plugin/utils/dojo.py
@@ -8,6 +8,8 @@ import functools
 import inspect
 import pathlib
 import urllib.request
+import base64
+import logging
 
 import yaml
 import requests
@@ -20,7 +22,7 @@ from CTFd.utils.user import get_current_user, is_admin
 
 from ..models import DojoAdmins, Dojos, DojoModules, DojoChallenges, DojoResources, DojoChallengeVisibilities, DojoResourceVisibilities, DojoModuleVisibilities
 from ..config import DOJOS_DIR
-from ..utils import get_current_container
+from ..utils import get_current_container, sanitize_survey
 
 
 DOJOS_TMP_DIR = DOJOS_DIR/"tmp"
@@ -69,6 +71,12 @@ DOJO_SPEC = Schema({
 
     Optional("auxiliary", default={}, ignore_extra_keys=True): dict,
 
+    Optional("survey"): {
+        Optional("probability"): float,
+        "prompt": str,
+        "data": str,
+    },
+
     Optional("modules", default=[]): [{
         **ID_NAME_DESCRIPTION,
         **VISIBILITY,
@@ -80,6 +88,12 @@ DOJO_SPEC = Schema({
         Optional("import"): {
             Optional("dojo"): UNIQUE_ID_REGEX,
             "module": ID_REGEX,
+        },
+
+        Optional("survey"): {
+            Optional("probability"): float,
+            "prompt": str,
+            "data": str,
         },
 
         Optional("challenges", default=[]): [{
@@ -102,6 +116,12 @@ DOJO_SPEC = Schema({
                 Optional("dojo"): UNIQUE_ID_REGEX,
                 Optional("module"): ID_REGEX,
                 "challenge": ID_REGEX,
+            },
+
+            Optional("survey"): {
+                Optional("probability"): float,
+                "prompt": str,
+                "data": str,
             },
         }],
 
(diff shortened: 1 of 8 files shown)


2stinkysocks:
Current to-do list:

- [x] HTML/CSS sanitization
- [x] Prevent survey styles from affecting rest of site
- [x] HTML Form implementation
- [x] One-click submit js implementation
- [x] CSRF prevention for survey submissions
- [x] Tested thumbs up/down surveys
- [x] Tested free-response surveys
- [x] Tested multiple choice surveys
- [x] Updating dojo shouldn't clear survey responses
- [x] Tested possible exploits
- [x] Verified non-survey challenges are unaffected
- [x] Better/more polished method for inputting survey data

Note: I removed the foreign key constraints from the survey responses table since I think there's no situation where we'd ever want to modify the data that's already in the table. Queries to this table can be done by manually checking if challenge/dojo/user match. This fixes the cascading deletes problem when updating a dojo.

2stinkysocks:
reminder to self to fix load_surveys so it loads the html into survey.src directly

2stinkysocks:
This branch should be ready for review now. I've tested existing dojos with no issues, and I've made multiple example surveys in my fork of example-dojo (https://github.com/2stinkysocks/example-dojo). These examples can probably added to documentation later. 

Surveys can be defined in yml by either providing a src (file path inside the directory specified by survey-sources), or data (the actual html data of the survey). 

Surveys are embedded into an html form that will make a request to the survey submit endpoint when the form is submitted, or alternatively submitted through a single click by specifying data-form-submit on a tag.

Currently surveys are stored in the db at a challenge-level (but can be defined in yml at challenge, module, or dojo level), and are sanitized before they are inserted into the db. Sanitization is done on a whitelist-basis; allowed html/css attributes can be found under sanitize_survey in dojo_plugin/utils/\_\_init\_\_.py.

There is currently no way to view survey responses in the frontend, it is just stored in the survey_responses table for now. 

I was unable to find any way to exploit this system with xss, I doubt this is possible but it would be helpful to have some external feedback.

ConnorNelson:
- NOTE: foreign keys deleted to handle the dojo delete/recreate problem when we do an update (which wipes everything out).

dojo_plugin/utils/dojo.py
```diff
@@ -143,24 +139,20 @@
...
+            },{
+                Optional("probability"): float,
+                "prompt": str,
+                "src": str,
+            },{
+                Optional("probability"): float,
+                "prompt": str,
+                "src": str,
+                "data": str
+            }),
```
ConnorNelson: Lets either validate before we further go to sources to build this thing up, or get rid of `src` entirely since it is irrelevant after.

dojo_plugin/utils/dojo.py
```diff
@@ -262,6 +254,30 @@ def load_dojo_subyamls(data, dojo_dir):
 
     return data
 
+def load_surveys(data, dojo_dir):
+    """
+    Optional survey data can be stored in an arbitrary directory under dojo_dir
+
+    This directory is specified by 'survey-sources' under the base yml file
+
+    This function copies the html survey data into the survey.src attribute
```
2stinkysocks: reminder to update this to survey.data

ConnorNelson:
Here is how this works (see https://github.com/2stinkysocks/example-dojo/tree/0bedca16286daed8a866f7d68d4b94aa46a52392):

- At the dojo level, we can optionally provide a `survey-sources` which is a path to a directory that will hold the surveys. There is no default value for this.
- At a dojo/module/challenge (any of them) level I can specify a `survey` field, which is a dictionary, that can hold either:
  - `survey-sources` based option (paths)
    - `prompt` (required) which is a human readable question being asked in the survey (e.g. not html)
    - `src` (required) which is a path relative to `survey-sources` containing html and css
    - `probability` (optional=1.0) which is a decimal value between 0 and 1.0, e.g. 0.5 means 50% chance of showing, 0.9 means 90% chance of showing.
  - inline option (e.g. not using `survey-sources`
    - `prompt` (required) which is a human readable question being asked in the survey (e.g. not html)
    - `data` (required) which is inline html/css
    - `probability` (optional=1.0) which is a decimal value between 0 and 1.0, e.g. 0.5 means 50% chance of showing, 0.9 means 90% chance of showing.

Right now, there is not an interface for downloading/viewing/exporting the responses (e.g. would have to manually get it from the database right now).

Issue #329, opened 2024-02-20
SSH: Support FIDO-based Login


adamdoupe:
Some users have requested to be able to use their FIDO/YUBI key ssh key.

```
sk-ecdsa-sha2-nistp256@openssh.com
sk-ssh-ed25519@openssh.com
```

Info is here (not sure if we need to enable/change the sshd config):
https://cryptsus.com/blog/how-to-configure-openssh-with-yubikey-security-keys-u2f-otp-authentication-ed25519-sk-ecdsa-sk-on-ubuntu-18.04.html

sjzhu:
Thanks adam! I think this regex would also need to be changed (because of the @openssh.com that's part of the protocol). https://github.com/pwncollege/dojo/blob/628f9eb36131ae6e100983e9d2d5e7026eaec66d/dojo_plugin/api/v1/ssh_key.py#L26

supercoolspy:
I tested it, and it seems like it should work without issues if we update the regex

ConnorNelson:
Solved by #715

PR #715, opened 2025-06-12
Site: Support more ssh key formats

supercoolspy:
Solves #329 and adds support beyond just regex for all ssh supported key formats, and expands on #559


supercoolspy:
Tested on my local installation and worked correctly

dojo_plugin/api/v1/ssh_key.py
```diff
@@ -23,17 +25,18 @@ def post(self):
...
+            try:
+                key = SSHKey(key_value, strict=True)
+                key.parse()
+                key_value = f"{key.key_type.decode()} {base64.b64encode(key._decoded_key).decode()}"
+            except (InvalidKeyError, NotImplementedError) as e:
                 return (
                     {
                         "success": False,
-                        "error": f"Invalid SSH Key, expected format:<br><code>{key_re}</code>"
+                        "error": f"Invalid SSH Key, error: <code>{e}</code> <br>Refer below for how to generate a valid ssh key"
```
spencerpogo, line 36: - are we certain that this error message either cannot be user controlled or gets sanitized? 
- worried that someone might edit the template to move/remove the "below" section and this could easily become outdated

dojo_plugin/api/v1/ssh_key.py
```diff
@@ -23,17 +25,18 @@ def post(self):
         key_value = data.get("ssh_key", "")
 
         if key_value:
-            key_re = "ssh-(rsa|ed25519|dss) AAAA[0-9A-Za-z+/]{1,730}[=]{0,2}"
-            key_match = re.match(key_re, key_value)
-            if not key_match:
+            try:
+                key = SSHKey(key_value, strict=True)
+                key.parse()
+                key_value = f"{key.key_type.decode()} {base64.b64encode(key._decoded_key).decode()}"
```
spencerpogo, line 31: noticed this in the `sshpubkeys` source:

```py
class SSHKey:
    # ...
  
    @property
    def key(self):
        """Base64 encoded key"""
        return base64.b64encode(self._decoded_key)
```

maybe this could do the same thing while being a less brittle against updates to the internals? `_` properities are usually not subject to semantic versioning

spencerpogo (commented):
couple nitpicks

dojo_plugin/api/v1/ssh_key.py
supercoolspy: > are we certain that this error message either cannot be user controlled or gets sanitized?

#742 good catch!

Issue #579, opened 2024-09-26
Why does `workspacefs` exist?


ConnorNelson:
Like seriously. The point is for the challenge to not be able to see into nix-land through the PATH, but why should a challenge *ever* trust the PATH? It's controlled by the user? If the challenge is going to do PATH resolution, it's going to do it on a trusted PATH, which just shouldn't include nix-land anyways. Why does this exist??

spencerpogo:
If the multi-purpose setuid interpreter handles this, then it should make it easy for challenge authors to do the right thing.

ConnorNelson:
So my best guess of what we were thinking a year ago, is that we don't want challenge binaries calling into nix in order to help prevent the separation. I'm not really sure where this happens, maybe a web security challenge reaching for firefox, or something like that. Probably the better answer is for challenges to use absolute paths when referencing other programs and/or for challenge programs to modify PATH to be what they want. There's a chance a challenge might more easily shoot themselves in the foot, but I'm not even totally sure about that. Either way, I don't think this is the right solution, and also #718 would have required a more-complicated path to continue supporting it (`virtiofs -> fusefs`, I think we'd always see `root:root` accessing the files, which would mean they would all be hidden).

And so, this is closed by #718, which removed workspacefs.

zardus:
I think resetting PATH should be in some set of best practices that we
maintain somewhere. As we realized a while back, if something is
nix-injectible, it's almost certainly also PATH-injectible.

PR #718, opened 2025-06-17
Workspace: Support host-isolated workspaces

ConnorNelson:


zardus:
What is this hipster stuff???

ConnorNelson:
Probably we don't actually want to just "always run kata", but instead we provide some interface for an image to say "hey i need a more-privileged context" (aka start a vm for me).

---

Before adding `/dev/shm` (specifically increasing its size), kata was just silently failing, and by mostly-luck I managed to find this was the issue and solution.

Now things boot, you can manually `docker exec` into the container, and stuff looks great (home directory is there, nix is there, etc).

The last remaining hold out on functionality seems to be networking.

---

https://github.com/kata-containers/kata-containers/blob/3.18.0/docs/design/architecture/networking.md talks about how networking works. Specifically, we should have a `tap0_kata` interface inside the container networking namespace (on the host). For some reason, it's not created.

If we run kata from the host (rather than inside our privileged infrastructure container), networking works great, `tap0_kata` is created.

I can even manually enter the network namespace and create a tap device. From there I'm not sure how to "hotplug" it into the vm. Qemu doesn't ever seem to be started with a flag to say "and here is my networking device", instead it seems to use this hotplug functionality that I am currently very-unfamiliar with.

In theory there's no reason we shouldn't be able to do this. Somewhere there is just a bug waiting to be understood. It could be as "simple" as something like `/dev/shm`, or maybe some weird thing where the network interface gets created but for some reason it appears in the root network namespace instead of the correct network namespace, or really who knows what.

ConnorNelson:
This might actually just be an issue/incompatibility with docker version 28.0.4.

ConnorNelson:
This PR now also addresses #579 by deleting workspacefs.

ConnorNelson:
Currently:
- Setting `cpu_period` and `cpu_quota` makes kata not-happy. I think the answer is to set `io.katacontainers.config.hypervisor.default_vcpus` as per https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/how-to-set-sandbox-config-kata.md.
- `/home/hacker` is not being mounted `nosuid`.

ConnorNelson:
There seems to be some sort of race condition surrounding getting output with `docker exec -i` and `kata`.

This is discussed in https://github.com/kata-containers/kata-containers/issues/9071.

ConnorNelson:
`Could not create the sandbox resource controller failed to write subtree controllers [cpu cpuset] to "/sys/fs/cgroup/cgroup.subtree_control": write /sys/fs/cgroup/cgroup.subtree_control: device or resource busy`

`sandbox_cgroup_only=true` in `/opt/kata/share/defaults/kata-containers/configuration.toml` fixes this, but it's also possible that the infrastructure container got into a weird state that is rare?

https://github.com/kata-containers/kata-containers/blob/main/docs/design/host-cgroups.md discusses this. It seems like maybe we would like `sandbox_cgroup_only=true` anyways to prevent a user from using excessive resources via overhead.

ConnorNelson:
Now we're just going to enable CAP_ADMIN if we're host-isolated, and remount /home/hacker nosuid ourselves.

It's not clear if we should be granting CAP_ADMIN, etc to these host-isolated workspaces or not. If we change our minds on this, we'll need another mechanism to make sure /home/hacker ends up nosuid.

This mechanism (and maybe also stuff like CAP_NET_ADMIN would be pretty cool for allowing a challenge to just do things without having to further start vms/namespace/etc though. We definitely can't trust the workspace's host though if we do this (or at best, it depends on the challenge).

zardus:
S E C U R I T Y

PR #721, opened 2025-06-22
Switch default dojo host

supercoolspy:


ConnorNelson:
It looks like Safari doesn't support this `*.localhost` behavior like everyone else. Maybe we should just say `localhost` or `127.0.0.1`?

dojo_plugin/api/v1/workspace.py
```diff
@@ -8,12 +11,14 @@
...
 from ...utils import get_current_container, container_password
 from ...utils.workspace import start_on_demand_service, reset_home
+from ...pages.workspace import forward_workspace
 
 
 workspace_namespace = Namespace(
     "workspace", description="Endpoint to manage workspace iframe urls"
 )
 
+signing_key = os.environ.get("WORKSPACE_SECRET")
```
ConnorNelson: Move this to `config.py` like the other environment variables. Then you'll just import `WORKSPACE_SECRET` from config, and use it in the relevant spots.

dojo_plugin/api/v1/workspace.py
```diff
@@ -53,11 +77,11 @@ def get(self):
                 "reconnect": 1,
                 "reconnect_delay": 200,
                 "resize": "remote",
-                "path": url_for("pwncollege_workspace.forward_workspace", service=service_param, service_path="websockify"),
+                "path": forward_workspace(service=service_param, service_path="websockify", sig=sig, container_id=container_id, include_host=False),
```
ConnorNelson: rename `sig` to `signature`.

dojo_plugin/api/v1/workspace.py
```diff
@@ -34,6 +39,25 @@ def get(self):
         if not container:
             return {"active": False}
 
+        if not signing_key:
+            abort(500)
+            return
+
+        container_id = container.id
+        if not container_id:
+            abort(400, "Invalid container")
```
ConnorNelson: Is this possible?

zardus:
Yo @supercoolspy, would you mind rebasing this on master? Let's start the journey of getting this merged :-)

supercoolspy:
Current issue is that with the new test changes it always uses the container's ip. The issue is that
A. It isn't localhost, so `localhost.pwn.college` won't work
B. No way to do ip subdomains

Working on setting a way to override the domain in local tester

zardus:
idea:

- move tests into a `tester` docker inside the outer docker
- add `test.pwn.college` and `workspace.test.college` into /etc/hosts
- ???
- profit!

nuggetpilot:
development in production = motivation speed run

supercoolspy:
> idea:
> 
>     * move tests into a `tester` docker inside the outer docker
> 
>     * add `test.pwn.college` and `workspace.test.college` into /etc/hosts

Do you think I should do it in this pr?

zardus:
Actually, why don’t we:

- add a workspace.localhost.pwn.college DNS entry
- expose ports on the host during multinode testing

Thoughts?

supercoolspy:
I think that would work. How do you think we should add the dns entry since you can't modify a container's hosts (via docker, have to modify the file) after the first start? I was thinking either a python or bash script that we run on the host. For multinode it may be a little harder since we will probably have to restart the container or something like that

zardus:
No, I mean we should actually just add this dns entry for real

supercoolspy:
> No, I mean we should actually just add this dns entry for real

OK, that should work! My only worry is nginx no longer resolving traffic coming from that route, but I doubt that would happen.

supercoolspy:
@codex review

deploy.sh
```diff
@@ -221,6 +233,25 @@ fi
...
+if [ -z "$DOJO_HOST_CONFIG" ]; then
+        DOJO_HOST_CONFIG="localhost.pwn.college"
+fi
+
+if [ -z "$WORKSPACE_HOST_CONFIG" ]; then
+        WORKSPACE_HOST_CONFIG="workspace.${DOJO_HOST_CONFIG}"
+fi
+
+if [[ "$DOJO_HOST_CONFIG" != "$CONTAINER_IP" && "$DOJO_HOST_CONFIG" =~ [A-Za-z] ]]; then
+        TEST_CONTAINER_EXTRA_ARGS+=("--add-host" "${DOJO_HOST_CONFIG}:${CONTAINER_IP}")
```
ConnorNelson, line 248: This helps make DNS work for testing, so tests can resolve IPs. This is for the main node.

deploy.sh
```diff
@@ -221,6 +233,25 @@ fi
...
+fi
+
+if [[ "$WORKSPACE_HOST_CONFIG" != "$CONTAINER_IP" && "$WORKSPACE_HOST_CONFIG" != "$DOJO_HOST_CONFIG" && "$WORKSPACE_HOST_CONFIG" =~ [A-Za-z] ]]; then
+        TEST_CONTAINER_EXTRA_ARGS+=("--add-host" "${WORKSPACE_HOST_CONFIG}:${CONTAINER_IP}")
```
ConnorNelson, line 252: And this is for the workspace node.

deploy.sh
```diff
@@ -268,6 +304,8 @@ if [ "$START" == "yes" -a "$MULTINODE" == "yes" ]; then
 	sleep 5
 	docker exec "$DOJO_CONTAINER" dojo compose restart ctfd sshd
 	sleep 5
+	docker exec "$DOJO_CONTAINER" dojo compose restart nginx
```
ConnorNelson, line 307: This fixes issue with master where sometimes we randomly fail multinode (timeout).

docker-compose.yml
```diff
@@ -241,12 +244,38 @@ services:
...
+      workspace_net:
+        aliases:
+          - nginx
+        ipv4_address: 10.0.0.3
+
+  nginx-workspace:
+    container_name: nginx-workspace
+    hostname: nginx-workspace
+    profiles:
+      - workspace_exclusive
```
ConnorNelson, line 262: New profile. Previously we had profiles for only main node, and then also one for everything. This now gives us a profile for only workspace.

dojo/dojo-init
```diff
@@ -36,9 +36,11 @@ define () {
 DEFAULT_DOJO_HOST=localhost.pwn.college
 
 define DOJO_HOST "${DEFAULT_DOJO_HOST}"
+define WORKSPACE_HOST workspace."${DEFAULT_DOJO_HOST}"
 define DOJO_ENV development
 define DOJO_WORKSPACE core
 define WORKSPACE_KEY
+define WORKSPACE_SECRET $(openssl rand -hex 16)
```
ConnorNelson, line 43: This secret is for urls, whereas the other one is for wireguard.

dojo/dojo
```diff
@@ -73,6 +73,7 @@ case "$ACTION" in
         multi_node=$(jq 'length > 0' /data/workspace_nodes.json 2>/dev/null || echo "unknown")
         if [ "$WORKSPACE_NODE" -gt 0 ]; then
             profiles+=("workspace")
+            profiles+=("workspace_exclusive")
```
ConnorNelson, line 76: On production migration, we need to make sure the new nginx service starts up, and then also that port 80 and 443 are open.

dojo/dojo-init
ConnorNelson: Also for production migration, we need to make sure we generate these secrets, and that everyone shares it.

dojo_theme/templates/workspace_port.html
```diff
@@ -0,0 +1,72 @@
...
+    name="{{ iframe_name }}"
+    src=""
+    class="d-none"
+></iframe>
+<h1 class="inactive">Loading...</h1>
+
+{% endblock %} {% block scripts %} {{ super() }}
+
+<script>
+    let response = fetch("/pwncollege_api/v1/workspace?port={{ port }}")
```
ConnorNelson, line 48: Realistically this template could be merged with workspace_service.html, since `?port=` vs `?service=` is the only difference.

nginx-workspace/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ADD https://github.com/nginx-modules/ngx_http_hmac_secure_link_module.git#master /ngx_http_hmac_secure_link_module
+
+WORKDIR /nginx
+
+RUN export LIBCLANG_PATH="$(llvm-config --libdir)" \
+    && auto/configure \
+    --with-compat \
+    --with-http_ssl_module \
+    --add-dynamic-module=/nginx-acme \
+    --add-dynamic-module=/ngx_http_hmac_secure_link_module \
```
ConnorNelson, line 29: We rebuild so that we can get `ngx_http_hmac_secure_link_module`.

nginx-workspace/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git build-essential ca-certificates \
+    libpcre2-dev libssl-dev zlib1g-dev \
+    curl pkg-config cmake \
+    clang llvm-dev libclang-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install rust for nginx
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y \
```
ConnorNelson, line 13: Base on nginx, and bring in rust, rather than current master's primary nginx, which bases on rust. This solves some sort of libc incompatibility issue while compiling.

docker-compose.yml
```diff
@@ -241,12 +244,38 @@ services:
...
+    restart: always
+    build: ./nginx-workspace
+    ports:
+      - "80:80"
+      - "443:443"
+      - "8888:8888"
```
ConnorNelson, line 268: This opens up port 8888, so that wireguard network can reach it.

nginx/templates/workspace.conf.template
```diff
@@ -0,0 +1,72 @@
+server_name $WORKSPACE_HOST;
+
+resolver 127.0.0.11 valid=30s ipv6=off;
+resolver_timeout 1s;
+
+location ~ ^/workspace/(?P<container_id>[^/:]+)/(?P<sig>[^/]+)/(?P<port>\d+)(?:/(?P<path>.*))?$ {
```
ConnorNelson, line 6: This guy is for single node.

nginx/templates/workspace.conf.template
```diff
@@ -0,0 +1,72 @@
...
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+
+    proxy_intercept_errors on;
+    error_page 502 = @workspace_not_found;
+
+    proxy_pass http://$container_id:$port/$path$is_args$args;
+}
+
+location ~ ^/workspace/(?P<container_id>[^/]+):(?P<target_host>[^/]+)/(?P<sig>[^/]+)/(?P<port>\d+)(?:/(?P<path>.*))?$ {
```
ConnorNelson, line 36: This guy is for multinode.

nginx-workspace/templates/route-redirecter.conf.template
```diff
@@ -0,0 +1,8 @@
+server {
+    server_name _;
+    listen 8888;
+
+    location ~ ^/workspace/(?P<container_id>[^/]+):(?P<target_host>[^/]+)/(?P<sig>[^/]+)/(?P<port>\d+)(?:/(?P<path>.*))?$ {
```
ConnorNelson, line 5: The purpose of `target_host` being in url path and doing this 8888 redirect stuff is so that the main node never actually needs to know the public-facing url of the workspace nodes.

dojo/dojo-init
```diff
@@ -36,9 +36,11 @@ define () {
 DEFAULT_DOJO_HOST=localhost.pwn.college
 
 define DOJO_HOST "${DEFAULT_DOJO_HOST}"
+define WORKSPACE_HOST workspace."${DEFAULT_DOJO_HOST}"
```
ConnorNelson, line 39: This is also super critical for a production migration, it ultimately gets used for ssl cert grabbing.

ConnorNelson:
Migration to production notes:
- Update `WORKSPACE_HOST` for each workspace node (e.g. `asu-1.workspace.pwn.college`, etc).
- Same `WORKSPACE_SECRET` on every node (including main): `openssl rand -hex 16`
- `git pull` everyone
- `dojo compose up -d --build nginx-workspace` on all workspace nodes. They should all get certs automatically now, and have services running on port 8888, 80, 443.
- `dojo compose up -d --build nginx ctfd` on the main node.

---

One potential hiccup will be if the workspace nodes infrastructure containers do not currently have port 80/443 exposed. Also we need to do cloudflare stuff (proxy through cloudflare, block everyone except cloudflare from talking to 443).

PR #722, opened 2025-06-24
Site: Fix practice mode in navbar

TheodorKitzenmaier:
Fixed an issue where restarting a challenge in practice mode using the navbar would start the challenge in normal mode instead of practice. Restarting will now preserve the current mode.

### DIFF
--- a/dojo_plugin/api/v1/docker.py
+++ b/dojo_plugin/api/v1/docker.py
@@ -407,9 +407,6 @@ class RunDocker(Resource):
             return {"success": False, "error": "No active challenge"}
 
         user = get_current_user()
-        if not user:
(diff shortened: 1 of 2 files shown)


dojo_plugin/api/v1/docker.py
```diff
@@ -404,9 +405,21 @@ def get(self):
         dojo_challenge = get_current_dojo_challenge()
         if not dojo_challenge:
             return {"success": False, "error": "No active challenge"}
+
+        user = get_current_user()
+        if not user:
+            return {"success": False, "error": "User not logged in"} #?
```
ConnorNelson: `@authed_only` should already guarantee this is true.

PR #724, opened 2025-06-27
Workspace: Fix nix gdb symbols

### LINKED ISSUE #467: Fix GDB Symbols
Nix is tricky! It seems like the challenge image might need to be a bit more explicit about things.

This shows that image gdb knows where to find libc debug info:
```
strace /usr/bin/gdb --batch -ex 'starti' -ex 'c' -ex 'p __libc_start_main' /bin/false 2>&1 | grep open | grep debug | grep 02430
```

Unfortunately nix gdb does not:
```
strace gdb --batch -ex 'starti' -ex 'c' -ex 'p __libc_start_main' /bin/false 2>&1 | grep open | grep debug | grep 02430
```

With `-ex 'set debug-file-directory /lib/debug'` nix gdb can find the debug info:
```
strace gdb --batch -ex 'set debug-file-directory /lib/debug' -ex 'starti' -ex 'c' -ex 'p __libc_start_main' /bin/false 2>&1 | grep open | grep debug | grep 02430
```

Rather than have users manually perform this, I think we can get a system gdbinit that always runs, or maybe theres an ENV variable for us to set, or otherwise we can patch nix gdb (but this sounds unideal because this is a *challenge image* attribute).

zrayaz1:
Adds a system gdbinit so nix gdb knows where to find debug symbols.

resolves #467

### DIFF
--- a/dojo_plugin/api/v1/dojos.py
+++ b/dojo_plugin/api/v1/dojos.py
@@ -142,92 +142,6 @@ class DojoSolveList(Resource):
         return {"success": True, "solves": solves}
 
 
-@dojos_namespace.route("/<dojo>/challenges/solve")
-class DojoChallengeSolve(Resource):
-    @authed_only
-    @dojo_route
-    def post(self, dojo):
-        user = get_current_user()
-        data = request.get_json()
-        dojo_challenge = (DojoChallenges.from_id(dojo.reference_id, data.get("module_id"), data.get("challenge_id"))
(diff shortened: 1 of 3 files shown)


ConnorNelson:
This will definitely work for an ubuntu based image, but if we have issues like #467 popping up, probably it's because either the image doesn't have the relevant debug symbols (e.g. for libc or whatever library), or those debug symbols aren't in `/lib/debug`.

PR #732, opened 2025-07-08
Share button changes to make it more usable

IronCobra22:


IronCobra22:
Forgot to write the change description so I'll leave it here

- removed the partially functioning share button and replaced it with a share message
- now has a box with the link to share and a copy button that copies the link

zardus:
Thanks!

PR #734, opened 2025-07-08
Update nix to 25.05

zrayaz1:
Pwndbg was removed from nixpkgs in 25.05 so instead referencing their flake.

Using ida-free version 8 from nix 24.11
ida-free moves to version 9 in 25.05 which no longer serves the binary due to a change in license.

### DIFF
--- a/workspace/Dockerfile
+++ b/workspace/Dockerfile
@@ -1,7 +1,7 @@
 FROM alpine
 
 RUN apk add --no-cache nix
-
+RUN apk add --no-cache git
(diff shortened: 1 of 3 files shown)


ConnorNelson:
`apk add --no-cache git` is needed for `pwndbg`, which fails because `git` is missing (I think it's related to https://github.com/NixOS/nix/issues/3533).

This seems like a bug in nix, but who knows, maybe it somehow makes sense in a nonobvious way?

PR #748, opened 2025-07-22
update seccomp profile location

zardus:
The moby project moved the seccomp profile to a different repo about two hours ago (https://github.com/moby/moby/commit/a600da91f47c2b2c8e2000cb0c5eb4992a289f61), which broke our builds. This updates the path to the new repo.


zardus:
(I confirmed that the files are identical)

PR #754, opened 2025-07-24
bring back the GOOD panel!

zardus:


zardus:
This panel cost us $7

PR #755, opened 2025-07-25
Allow specifying challenges as a resource type, ordered in with the other resources.

zardus:
We've discussed this for a while, and it's here. This one was pricy:

```
Total cost:            $92.39
Total duration (API):  1h 49m 18.4s
Total duration (wall): 6h 57m 46.3s
Total code changes:    902 lines added, 416 lines removed
Usage by model:
    claude-3-5-haiku:  175.8k input, 7.2k output, 0 cache read, 0 cache write
         claude-opus:  1.2k input, 96.1k output, 32.3m cache read, 1.9m cache write
```

We should heavily review the code, because claude did this almost autonomously. Theoretically, it claims that:

- you can now add challenges as any other resource (`type: challenge`).
- you can still add challenges the old fashioned way (they'll just be added at the end)

We might want to add a `type: header` resource that recreates an `<h2>` separator.

### DIFF
--- a/dojo_plugin/models/__init__.py
+++ b/dojo_plugin/models/__init__.py
@@ -422,7 +422,8 @@ class DojoModules(db.Model):
     @delete_before_insert("_resources")
     def resources(self, value):
         for resource_index, resource in enumerate(value):
-            resource.resource_index = resource_index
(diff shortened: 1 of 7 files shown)


zardus:
yolo

PR #756, opened 2025-07-25
add a groundrules nag screen to make sure people see them

zardus:
People keep missing our groundrules somehow. This adds a nag screen that (once per browser) will have people type a commitment not to share writeups.
image


zardus:
@robwaz you have :confused: emoji?

PR #760, opened 2025-07-29
support defaults for dojos with a nulled data attribute

zardus:
fixes missing dojo scoreboard on some dojos after the hidden-scoreboard feature merge


zardus:
fingers crossed

PR #771, opened 2025-08-04
UX Improvements

TheodorKitzenmaier:
Created a new branch focused on streamlining the user experience.

Some planned features are:
-	Starting the next challenge automatically after finishing the current challenge.
-	Scrolling to the next (unsolved?) challenge in the modules page after finishing the current challenge.
-	Control menu to toggle these behaviors.
-	Suggestions welcome!

### DIFF
--- a/dojo_theme/templates/workspace.html
+++ b/dojo_theme/templates/workspace.html
@@ -5,11 +5,17 @@
 {{ super() }}
 
 <style>
+    :root {
+        --background: #111;
+        --foreground: #222;
+        --outline: #6c757d;
+    }
+
     html, body {
         margin: 0;
         height: 100%;
         min-height: 100%;
-        background-color: #111;
+        background-color: var(--background);
     }
 
     body {
@@ -46,11 +52,12 @@
     .workspace-control {
         height: 100%;
         margin: 0 2px;
-        border: 0;
+        border-width: 0px;
         border-radius: 1em;
         appearance: none;
         -webkit-appearance: none;
         -moz-appearance: none;
+        background-color: var(--foreground);
     }
 
     button.workspace-control  {
@@ -65,7 +72,7 @@
         box-sizing: border-box;
         padding: 1em;
         color: #fff;
-        background-color: #111;
+        background-color: var(--foreground);
         border: 0;
         border-color: #000 !important;
         border-radius: 1em;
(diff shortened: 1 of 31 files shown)


TheodorKitzenmaier:
Make it more clear that code/desktop is a selector using a hamburger.

Improve mobile browser support:
- SSH option.
- For small screens/windows, replace the workspace iframe with a link to the workspace.

TheodorKitzenmaier:
Invalid flag format notification.

zardus:
More wishlist:

- tweak the CSS (thin outlines for the input elements, vertically centering "Start" and "Practice" within the button)
- look into making the reset buttons more clear (right now, i see them as "lock" and "unlock" rather than "reset" and "reset in practice"
- we changed the "Practice" terminology to "Privileged". Is this intended?
- add a "None" workspace option that doesn't take up screen real-estate, for the true sshers
- hamburger menu icon in the workspace selector to make it more clear that it's a menu

zardus:
We should consider an explicit submit button for people that are used to not including `pwn.college{}`

dojo_theme/templates/workspace.html
```diff
@@ -105,26 +135,39 @@
...
-    <button id="start-unprivileged" title="Restart unprivileged" class="workspace-control btn btn-md btn-outline-secondary btn-challenge-start">
+    <script>var practice = true;</script>
+    {% else %}
+    <button id="challenge-switch" title="Restart privileged" class="workspace-control btn btn-md btn-outline-secondary btn-challenge-start">
         <i class="fas fa-lock"></i>
     </button>
+    <script>var practice = false;</script>
+    {% endif %}
+    {% else %}
+    <script>var practice = false;</script>
```
ConnorNelson: Consider something like:
```j2
<button id="challenge-switch"
        class="workspace-control btn btn-md btn-outline-secondary btn-challenge-start"
        title="Restart {{ 'unprivileged' if privileged else 'privileged' }}"
        data-privileged="{{ 'true' if privileged else 'false' }}">
    <i class="fas {{ 'fa-unlock' if privileged else 'fa-lock' }}"></i>
</button>
```

dojo_theme/templates/workspace.html
```diff
@@ -65,21 +72,44 @@
...
         opacity: 1;
     }
 
     .workspace-control input:focus:not(.submit-correct):not(.submit-incorrect):not(.submit-warn),
     .workspace-control input:hover:not(.submit-correct):not(.submit-incorrect):not(.submit-warn) {
-        outline: 2px solid #6c757d !important;
+        outline: 0 !important;
+    }
+
+    .workspace-selector {
```
ConnorNelson: Should this just be `.workspace-control select`?

dojo_theme/templates/workspace.html
```diff
@@ -65,21 +72,44 @@
...
+        height: 100%;
+        border: 0;
+        background-color: var(--foreground);
+        color: var(--outline);
+        font-size: 16px;
+        line-height: 20px;
+        font-weight: 700;
+    }
+
+    .workspace-highlight {
```
ConnorNelson, line 105: Should this just be `.workspace-control i`?

TheodorKitzenmaier: This seems to mess with some of the other icons, which end up disappearing when buttons are hovered.

PR #783, opened 2025-08-08
UX Improvements Round 2

TheodorKitzenmaier:
# Current Planned UX Improvements:
- Vertically center start text (done).
- Add "none" workspace content option, which shrinks the workspace iframe to just include the action bar.
- Option to disable auto-submit.
- Indication of when auto-submission is triggered.
- Option to scroll to / start the next challenge after finishing the current challenge.


ConnorNelson:
Instead of none let’s call it SSH, and make it even just be 1 sentence with a link to setting page for how to configure ssh. I think it solves the same problem our ssh users are interested in (minimal popup) and makes more sense to reveal our final way of accessing the workspace.

TheodorKitzenmaier:
Menu doesn't play nice with the tiny SSH window, but I don't have enough brain cells floating around to fix it tonight.

Issue #640, opened 2025-01-18
Delay in updating the correct submission


GoelUmang:
When submitted the flag, it processes the flag slow, thus delaying in correctness. Could be solved by reloading the page

robwaz:
Issue should be addressed with PR referenced above.

PR #793, opened 2025-08-11
Refresh flag icon when an active challenge is solved

sul3chu:
I added a empty function for `updateNavbarDropdown` because I saw an error in the console at pwn.college.
image

We can remove it if it’s an old function or something planned for later.

However, the main purpose of this PR is to refresh the flag icon when a flag is successfully submitted.

Accessing the iframe’s parent content works as long as the iframe and the parent page share the same origin.
I mention this because in production they might be on different domains, subdomains, or origins, which would block direct access.

### DIFF
--- a/dojo_theme/static/js/dojo/challenges.js
+++ b/dojo_theme/static/js/dojo/challenges.js
@@ -424,6 +424,11 @@ function windowResizeCallback(event) {
     $(".challenge-iframe").not(".challenge-iframe-fs").css("aspect-ratio", `${window.innerWidth} / ${window.innerHeight}`);
 }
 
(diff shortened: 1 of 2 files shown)


robwaz:
The `updateNavbarDropdown` function was removed in a recent UI change.  If you could remove

```
setTimeout(() => updateNavbarDropdown(), 1000);
```

and the stub function from this PR we can merge it.

sul3chu:
@robwaz I removed it. Thanks. 

I think this PR solve issue #640, I don't know how to relate/link it. I just read it and think that issue is talking exactly what we are improving with these changes.

robwaz:
Closes #640

PR #802, opened 2025-08-15
more testing tweaks

zardus:


zardus:
CI says we're running out of space, but it makes no sense...

PR #812, opened 2025-08-16
Site: Add SSH workspace option

TheodorKitzenmaier:
Adds a SSH workspace option which minimizes the amount of space taken up on the modules page.


TheodorKitzenmaier:
For now SSH is only in the modules page since it doesn't make a ton of sense to me for it to also be in the dedicated workspace. It could easily be added to the workspace by adding a hint and SSH to the options in workspace.py if needed.

PR #819, opened 2025-08-18
Actionbar: Make the entire selector box clickable

### LINKED ISSUE #818: hamburger should be clickable
The hamburger menu on the workspace switcher isn't clickable --- the click is only registered on the text. Can we make that whole button area clickable?

TheodorKitzenmaier:
This PR revises the workspace selector so that the burger menu also opens the workspace selection menu.

This is achieved by creating a custom html template which uses buttons to replicate the behavior of the normal <select> element. A fun side effect is that we can now style this menu!

closes #818


TheodorKitzenmaier:
Time for a less janky solution.

PR #826, opened 2025-08-19
try to control disk space with btrfs COW subvolumes

zardus:


zardus:
this one might have done the trick

Issue #827, opened 2025-08-19
Add `interface` field for challenges


ConnorNelson:
Something like this:
```
challenges:
- id: test
  description: Wow this is a test challenge
  interface:
    Code: 8401
    SSH: 'ssh'
```

That shows the ssh special case.

We should track a list of most recently used (interface_name, port) pairs, and walk that list for what to default to otherwise default to the first one. This is all client side for determine this.

robwaz:
Feature added in #828

PR #828, opened 2025-08-19
Site: Add challenge interface configuration

TheodorKitzenmaier:
This PR sets the groundwork for challenges which are able to specify alternate ways with interacting with them by adding an optional interface parameter for the challenge/module/dojo.


TheodorKitzenmaier:
closes #827

zardus:
shouldn't it be "interfaces" rather than "interface" in the yml?

Also, why do we need a new endpoint? Can't this be passed along by the `modules.html` and `workspace.html` templates?

PR #829, opened 2025-08-20
feat: Added Latex Support

Ahmahsamara17:
added MathJax to base.html so that challenge descriptions and pages in Dojo can display LaTeX math expressions properly instead of raw $...$ text.

### DIFF
--- a/dojo_theme/templates/base.html
+++ b/dojo_theme/templates/base.html
@@ -102,4 +102,13 @@
 
     {{ Configs.theme_footer }}
   </body>
+<script>
+  window.MathJax = {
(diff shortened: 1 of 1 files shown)


ConnorNelson:
```
<script>
  window.MathJax = {
    tex: { inlineMath: [['$','$'], ['\\(','\\)']], displayMath: [['$$','$$'], ['\\[','\\]']] },
    options: {
      ignoreHtmlClass: ".*",         // ignore everything
      processHtmlClass: "math"       // but DO process inside .math
    }
  };
</script>
<script async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml.js"></script>
```

I think an opt in approach like this would make more sense, so we don’t suddenly start doing weird things with $. We would also need to update the markdown render whitelist to allow the math class through.

zardus:
:eyes:

PR #830, opened 2025-08-20
Site: Use Interface for Workspace Action Bar

TheodorKitzenmaier:
Adjusts the actionbar to work with the changes proposed in #828.


TheodorKitzenmaier:
Also implements the workspace history as described in #827.

PR #834, opened 2025-08-21
stats.py: Remove an irrelevant and buggy line introduced in 52a51cb.

ltfish:
This line uses `module`, which is not defined anywhere in this method, and writes to `challenge_solves`, which is not used anywhere in this method. It is likely an irrelevant line that @zardus accidentally committed from his local working copy.


ConnorNelson:
Thanks!

PR #837, opened 2025-08-21
Infrastructure: Apply CTFd patches

ConnorNelson:


ConnorNelson:
As part of the migration, we will run:
```sql
DROP TABLE solves;

CREATE UNIQUE INDEX solves_challenge_id_user_id_key
ON submissions (challenge_id, user_id)
WHERE type = 'correct' AND user_id IS NOT NULL;

CREATE UNIQUE INDEX solves_challenge_id_team_id_key
ON submissions (challenge_id, team_id)
WHERE type = 'correct' AND team_id IS NOT NULL;
```

Issue #838, opened 2025-08-21
Workspace does not support low width


TheodorKitzenmaier:
When viewed with a low window width, the challenge controls spill out of the actionbar.

<img width="604" height="771" alt="Image" src="https://github.com/user-attachments/assets/3504cb6e-b764-40e7-8170-6b303632238f" />

PR #839, opened 2025-08-22
Flex the actionbar

TheodorKitzenmaier:
Actionbar elements will now not spill out when the window is narrow. Items will move to a second line.


TheodorKitzenmaier:
closes #838

PR #840, opened 2025-08-22
Site: Fix ordered interfaces

TheodorKitzenmaier:
Interfaces will now be shown in the order they are specified in the dojo's yml. This also gives better control of what interface will be shown to a new user, as which interface is first is now predictable.

Interfaces now specify a name and port; ssh is still a special case (no port).
```
name: Module Name
challenges:
  - id: demoid
    name: Demo Challenge Name
    interfaces:
    - name: SSH
    - name: Terminal
      7681
    - name: Code
      port: 8080
```


ConnorNelson:
As a note, after we deploy this, we're going to need to use `dojo flask` to migrate any existing challenge interfaces to this new format. Then we will need to notify people using interfaces in their own dojo, and also update all of our dojos.

ConnorNelson:
@TheodorKitzenmaier can you make sure all tests work, except I think the welcome dojo challenges which use the old interface style until we merge.


The CI should be working again now, except for that issue which is PR specific.

TheodorKitzenmaier:
Since the example dojo was also updated with an example of the lecture interface, almost all of the tests are broken since the testing dojos import from the example dojo.

PR #849, opened 2025-08-23
Catch workspace 404s

### LINKED ISSUE #848: 404s in Workspace Interface
We can get a 404 workspace interface. This seems to stem from the following chain of events:

- `GET /pwncollege_api/v1/workspace?service=code` returns a `500`
- `actionbar.js` result["iframe_src"] is undefined, and so the iframe gets a url with the literal string "undefined" : https://github.com/pwncollege/dojo/blob/e6d17d1fd40b45140acff63a7cf32c9d8ff50841/dojo_theme/static/js/dojo/actionbar.js#L83
- That page does not exist, and so 404.

The `500` stems from this:
```
  File "/opt/CTFd/CTFd/plugins/dojo_plugin/api/v1/workspace.py", line 76, in get
    if start_on_demand_service(user, service) is False:
       ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/opt/CTFd/CTFd/plugins/dojo_plugin/utils/workspace.py", line 12, in start_on_demand_service
    exec_run(
    ~~~~~~~~^
        f"/run/current-system/sw/bin/dojo-{service_name}",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        assert_success=True,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/CTFd/CTFd/plugins/dojo_plugin/utils/workspace.py", line 40, in exec_run
    assert exit_code in (0, None), output
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError: b''
```

And so the issue is twofold. We should not be returning a 500 in this api endpoint. Fair enough if the service fails to start (really not sure why...) but we should communicate the error in our json response. For example, `{"success": False, "error": f"Service failed to start"}`, and then giving an interface indication of this, maybe it should be like the `SSH` logic, since maybe a different interface (if switched too) won't fail to start.

TheodorKitzenmaier:
Backend now catches the assertion which is raised when services fail to start and returns `success: False`.

Frontend now checks if starting the service failed, and displays a message accordingly.

closes #848

### DIFF
--- a/dojo_plugin/utils/workspace.py
+++ b/dojo_plugin/utils/workspace.py
@@ -7,7 +7,7 @@ on_demand_services = { "terminal", "code", "desktop"}
 
 def start_on_demand_service(user, service_name):
     if service_name not in on_demand_services:
-        return False
+        return
     try:
         exec_run(
             f"/run/current-system/sw/bin/dojo-{service_name}",


dojo_plugin/utils/workspace.py
```diff
@@ -7,15 +7,15 @@
 
 def start_on_demand_service(user, service_name):
     if service_name not in on_demand_services:
-        return
+        return False
```
zardus: Won't this break any service that isn't a start_on_demand service? Do the lectures still work with this line?

zardus:
:+1:

PR #852, opened 2025-08-23
add a testcase for lecture challenges

zardus:


zardus:
That vibecheck result is some hallucinated shit

PR #856, opened 2025-08-24
Reduce the fog of war

zardus:
This PR aims to make the dojo more understandable from the outside. Right now, it configures nginx to pass its request_id as a PWN-Trace-ID header to handlers, and modifies the ctfd container to log this header in all log output.

Stuff to review specifically:
- we do some dirty hooking.
- this will change the log output of everything in the ctfd container

Next steps:
- see about passing this through to homefs (might be tricky through docker, but should be doable)
- look into having stuff in the ctfd container also log the user id and such


zardus:
homefs is not easy: docker is not designed to log a custom http header all the way through its processing. Merging this as is for now; will think about homefs later

PR #857, opened 2025-08-24
Handle 404 or other errors for custom interfaces

TheodorKitzenmaier:
This PR extends some of the logic for handling failed connections to services to custom interfaces. The user will not see a 404, and instead is told to try reloading the challenge or contact the dojo administrator.


ConnorNelson:
@TheodorKitzenmaier can you refactor it so it's not recursive.

Issue #859, opened 2025-08-24
[Bug] Module layout display order


robwaz:
There appears to be a bug in how the module page is rendered.  Likely having something to do with importing.

Example case:

This is a module displaying as intended
  https://pwn.college/program-security/program-security/

This is the same module being imported as a module.
  https://pwn.college/cse466-f2025/program-security/

PR #860, opened 2025-08-25
keep resource_index when using imports

2stinkysocks:
Fixes https://github.com/pwncollege/dojo/issues/859, the dojo import process passes in the resource_index to DojoResources but it's never used, so it ends up with the wrong value and resources are displayed out of order.


zardus:
Good find!

PR #861, opened 2025-08-25
fix multinode tests by downgrading docker

zardus:


zardus:
@ConnorNelson should we merge this?

zardus:
Merging this before someone upgrades our poor nodes.

PR #862, opened 2025-08-25
catch docker errors that propagate up via requests

zardus:


zardus:
I think with this, we should expect to have no uncaught exceptions in the daily report!

PR #866, opened 2025-08-25
Create a valid $XDG_RUNTIME_DIR

### LINKED ISSUE #865: $XDG_RUNTIME_DIR points to a non existent directory
Some programs need a valid `$XDG_RUNTIME_DIR` to work like firefox, tmux. Both of them work since they have fallback code to choose other dirs in case this case.

```
(firefox:1822053): dconf-CRITICAL **: 02:27:20.248: unable to create directory '/run/user/1000/dconf': Permission denied.  dconf will not work properly.
[Parent 1822053, Unnamed thread 7269cfa39a60] WARNING: unable to create directory '/run/user/1000/dconf': Permission denied.  dconf will not work properly.: 'glib warning', file /build/firefox-141.0/toolkit/xre/nsSigHandlers.cpp:201
```

But some other program may not have fallback code like that, and will fail to start, in example zellij (an alternative to tmux).

ptrca:
Fixes #865 

I'm not familiar with nix or the code base. So please forgive my mistakes.

### DIFF
--- a/workspace/core/init.nix
+++ b/workspace/core/init.nix
@@ -62,7 +62,7 @@ let
     fi
     echo $DOJO_FLAG | install -m 400 /dev/stdin /flag
 
-    for path in /home/hacker /home/hacker/.config; do
+    for path in /home/hacker /home/hacker/.config /run/user/1000; do
       test -L "$path" && rm -f "$path"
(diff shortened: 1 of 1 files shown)


ptrca:
This is wrong though. We need to create $XDG_RUNTIME_DIR before running the desktop service/environment. $XDG_RUNTIME_DIR is still needed by terminal session apps.

PR #867, opened 2025-08-25
Synchronize Challenge Workspaces

TheodorKitzenmaier:
The PR aims to synchronize which challenge is being worked on between all open workspaces.

Currently, the workspace loads information about a challenge when the page is loaded. This means that even if a user starts another challenge, that workspace will retain the id, name, etc of the previous challenge. In order to use that workspace for the new challenge, the user needs to reload the page. Reloading the page also resets which interface is in use. If a user is using multiple workspace tabs (such as desktop + vscode), both will use the same interface when reloaded, instead of the interface last used in each respective workspace tab.

This can be solved by dispatching an event whenever a challenge is started, and having workspaces update themselves with the most up-to-date challenge information.

### DIFF
--- a/dojo_plugin/api/v1/workspace.py
+++ b/dojo_plugin/api/v1/workspace.py
@@ -6,6 +6,7 @@ from CTFd.models import Users
 from CTFd.utils.user import get_current_user, is_admin
 from CTFd.utils.decorators import authed_only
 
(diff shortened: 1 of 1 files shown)


TheodorKitzenmaier:
Moved to #891

PR #880, opened 2025-09-02
Site: Fix module import displaying challenges in wrong order

2stinkysocks:
Actually fixes the bug from the previous pr, I don't know why it worked for me last time but this seems to actually fix the issue. I'm just making sure unified_index is also copied over to the new challenges during the import, not just the resources.

This should fix the current live 466 dojo formatting being messed up (after a dojo update)


2stinkysocks:
This fixes module imports but breaks challenge imports, need to look into it more.

2stinkysocks:
Actually fixes https://github.com/pwncollege/dojo/issues/859 now

Issue #787, opened 2025-08-09
Restart Challenge button can target the wrong challenge


realdonkey:
With two windows open on pwncollege, the restart challenge button will restart the currently active challenge, instead of restarting the challenge which the button is attached to. 

Encountered when the user:
opens tab 1 on module A and starts challenge C1
opens tab 2 on module B and starts challenge C2
user returns to tab 1 and hits "restart challenge" on challenge C1

A terminal will be opened on tab 1 but the active challenge will be C2

ConnorNelson:
@TheodorKitzenmaier I think this makes sense because we talk to the api to determine our current challenge. Instead we need to embed the information into action bar and grab from there (`data-dojo-id="..."`, etc).

TheodorKitzenmaier:
I think the info is already embedded for submission, should be easy enough to hook the challenge starting into that as well.

PR #891, opened 2025-09-09
Site: Synchronize Active Challenge Between Workspace Tabs

TheodorKitzenmaier:
This PR aims to fix an issue where workspace tabs retain information about the challenge which was active when the tab was loaded. If the user opens a workspace tab, starts a new challenge, then tries to restart or submit using the old workspace tab, these actions will use the old challenge ID, despite not matching the current challenge.

In the past, this would be resolved by reloading your workspace tabs. However, this would reset all tabs to be using the same most recent workspace service.

This PR fixes this by having tabs communicate whenever a challenge is started. Workspace tabs in particular will update their contents to reflect information about the new challenge (ID, Name, Interfaces, etc.)

### DIFF
--- a/dojo_theme/static/js/dojo/actionbar.js
+++ b/dojo_theme/static/js/dojo/actionbar.js
@@ -49,7 +49,7 @@ function getRecentService(root) {
     return match;
 }
 
-function specialSelect(serviceName) {
+function specialSelect(serviceName, content) {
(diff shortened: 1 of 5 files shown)


TheodorKitzenmaier:
Hope I didn't miss anything, brain is a bit mushy at 2AM...

TheodorKitzenmaier:
closes #787

PR #896, opened 2025-09-10
Multiple dojos in one repository

2stinkysocks:
Adds base functionality for including multiple dojos in a single repository (CURRENTLY UNFINISHED).

As of now, the dojo.yml can optionally have a `dojos` key, which is a list of standard dojo yml objects. All modules live in the base directory, and are loaded conditionally per dojo yml object. (https://github.com/2stinkysocks/example-monorepo-dojo)

Current issues:
All dojos in a shared repository must not exist during dojo creation, or they will all fail.
There is no mechanism for migrating existing dojos into a single repository.
I have not tested a lot of existing features, and there may be some side effects I haven't found.
I have not tested github ci at all.

Currently working:
Shared repositories can be loaded as usual, multiple dojos will be created. 

Updating dojos will update only the one for which update was clicked. The files for shared repositories are updated any time any dojo is updated, but the database will only be changed for the dojo which update was called on (I don't think this has side effects, needs further testing). The git hash for all dojos in a single repo will be the same in the admin page.
Files for shared repos are stored under the same directory, with the format `shared-{repository_url_md5}`

Promoting dojos works (single boolean column change).

### DIFF
--- a/dojo_plugin/api/v1/dojos.py
+++ b/dojo_plugin/api/v1/dojos.py
@@ -91,12 +91,15 @@ class CreateDojo(Resource):
             return {"success": False, "error": "You can only create 1 dojo per day."}, 429
 
         try:
-            dojo = dojo_create(user, repository, public_key, private_key, spec)
+            dojos = dojo_create(user, repository, public_key, private_key, spec)
         except RuntimeError as e:
             return {"success": False, "error": str(e)}, 400
 
         cache.set(key, 1, timeout=timeout)
-        return {"success": True, "dojo": dojo.reference_id}
+        if len(dojos) == 1:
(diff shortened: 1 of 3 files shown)


2stinkysocks:
Note: going to change this functionality so creating a dojo has another extra field to specify a dojo yml path, so only one dojo will be created from the create page at a time

2stinkysocks:
New info:

Added an extra field in dojo create to specify a dojo yml path. Dojo creation only loads a single dojo which may be part of a larger monorepo. 

Repositories are duplicated on disk (one per dojo) and are updated when the specific dojo is updated.

ConnorNelson:
We want to make sure there's no symlink/path-traversal shenanigans.

PR #907, opened 2025-09-18
Improve error message for duplicate challenge IDs to include module id for easier debugging

prathamgupta36:
**Purpose of the PR:**

When parsing the yml files if a module contains duplicate challenge `ids` we now raise an assertion caused it to hit an `Internal Server Error`, with this PR that includes the module id:

```py
raise AssertionError(f"Duplicate challenge id: {challenge_id} in module {module_data.get('id')}")
```

This makes the error message explicit about *which module* contains the duplicate id, improving diagnosability compared to a generic "Internal Server Error".

**Why this change**

* Previously a duplicate challenge id could result in a vague server error that made it hard to identify the source of the problem.
* Returning a clear assertion message that includes the module id helps maintainers and dojo authors find and fix the problematic module quickly.

**Before**

* Duplicate challenge IDs → generic Internal Server Error (hard to trace).
error_before

**After**

* Duplicate challenge IDs → `AssertionError("Duplicate challenge id: <id> in module <module_id>")`, making it clear which module needs attention.
error_after

**Files changed**

* `dojo_plugin/utils/dojo.py`: inside `dojo_from_spec(...)` function, while iterating module resources.

**How to test**

1. Create a dojo with a `module.yml` where one module contains two resources of type `challenge` with the same `id`.
2. Click on the `update` button on admin panel the dojo.
3. Confirm the raised error message contains the duplicate challenge id and the module id:

```
AssertionError: Duplicate challenge id: <duplicate_id> in module <module_id>
```

**TLDR: clearer error when a module contains duplicate challenge IDs from generic `Internal Server Error` to message now includes the module id to help locate the issue.**

### DIFF
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -200,26 +200,32 @@ def update_dojo(dojo, update_code=None):
         db.session.rollback()
         error = str(e)
         match = re.search(r"Key \(dojo_id, module_index, id\)=\(.*?,\s*(\d+),\s*([^\)]+)\)", error)
-        if match:
-            module_index, challenge_id = match.groups()
-            module_index = int(module_index)
-            challenge_id = challenge_id.strip()
-            try:
-                module = dojo.modules[module_index]
-                challenge = next((c for c in module.challenges if c.id == challenge_id), None)
-                module_name = module.name
-                challenge_name = challenge.name if challenge else challenge_id
-                error_message = f"Duplicate ID used in Module: {module_name}, for Challenge: {challenge_name}."
-                return {"success": False, "error": error_message}, 400
-            except (IndexError, StopIteration):
-                 return {"success": False, "error": "Database integrity error: Could not parse module/challenge."}, 400
-        else:
-            print(f"ERROR: Dojo failed for {dojo}", file=sys.stderr, flush=True)
+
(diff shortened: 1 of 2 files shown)


ConnorNelson:
What is the current stack trace when this happens? Instead of managing a set (and in order to prevent races), lets let the db work for us. We should just catch the exception, and then reraise our exception with a better error message.

prathamgupta36:
Updated the implementation:
I've removed the manual tracking of challenge IDs with a set and now rely on the database unique constraint to detect duplicates.
The code now catches the `IntegrityError` during the database commit and re-raises a `RuntimeError` with a more informative error message that includes the module and challenge ID, for example:
```
RuntimeError: Duplicate challenge id: 'challenge-one' in module 'module-one'
```
This makes it much easier to diagnose duplicate challenge IDs without needing to inspect the stack trace.

prathamgupta36:
Working on making the error look better. This would fix #913.

prathamgupta36:
This fixes the issue without using `set()`, this does not change any other functionality.

prathamgupta36:
Reverted the changes made by the old approach. I have tested it, and this should now be ready to merge.

prathamgupta36:
Refactored the code to make it cleaner.

PR #912, opened 2025-09-19
Update ssh hint to dojo.pwn.college in the dojo_theme templates

maxtrap:
Changed the ssh hint from `hacker@pwn.college` to `hacker@dojo.pwn.college`.

@robwaz 

# Before
Screenshot from 2025-09-18 18-29-56
Screenshot from 2025-09-18 18-28-28

# After
Screenshot from 2025-09-18 18-29-39
Screenshot from 2025-09-18 18-29-00

### DIFF
--- a/dojo_theme/templates/module.html
+++ b/dojo_theme/templates/module.html
@@ -262,7 +262,7 @@
               </div>
               <div class="workspace-ssh">
                 <h3>Connect with SSH</h3>
-                Link your <a href="/settings#ssh-key" target="_blank">SSH key</a>, then connect with: <code>ssh hacker@pwn.college</code>
+                Link your <a href="/settings#ssh-key" target="_blank">SSH key</a>, then connect with: <code>ssh hacker@dojo.pwn.college</code>
               </div>
               {% include "components/actionbar.html" %}
             </div>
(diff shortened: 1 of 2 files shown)


robwaz:
Thanks!

zardus:
@ConnorNelson what do you think about making "ssh.pwn.college", instead?

zardus:
(I should clarify that my comment was orthogonal to this PR :rofl:)

ConnorNelson:
Definitely open to another domain. I think “workspace.pwn.college” might also make sense given our current terminology?Eventually when we get real fancy, everyone will probably have a dedicated hostname (per user name or something) so that we can route them straight to the node their workspace is active on. But we have time til that happens hahaha.On Sep 18, 2025, at 8:56 PM, Yan Shoshitaishvili ***@***.***> wrote:﻿zardus left a comment (pwncollege/dojo#912)
(I should clarify that my comment was orthogonal to this PR 🤣)

robwaz:
While we certainly can change the domain... I would vote for keeping `dojo.pwn.college` in addition to the chosen new domain for backwards compatibility.  There's likely hundreds of videos of `ssh hacker@dojo.pwn.college` being used.  I know it was my default for a long time on stream.  Removing it breaks user continuity.  We can certainly point learners in the getting started dojo to the new domain.

Issue #810, opened 2025-08-16
SSH no longer reconnects to newly started challenges


robwaz:
SSH sessions are not being handed off to new challenges.  Instead, the ssh session freezes and the user must reconnect.

To reproduce:
- Start a challenge
- ssh to dojo
- Start a different challenge

zardus:
I opened a #814 _just with a testcase_ that checks for the correct behavior (and currently fails). That testcase should eventually make it into master so that this doesn't happen again


For completion, a temporary workaround is `autossh`

ConnorNelson:
I'm not reproducing this.

Just started `Apple` on https://pwn.college/example/hello/. Then I ssh'd. Then I started `Banana` and it connected me just fine.

robwaz:
Status update.  This issue still does occur.  It is sporadic.  Maybe 1/4 times I change challenges the ssh does not handoff smoothly between challenges.

ConnorNelson:
@robwaz Can you confirm this happens on all challenges sporadically? As far as I can tell it happens every time on a kata challenge.

ConnorNelson:
Simplified debug loop:

```
$ docker run -d --rm --runtime=io.containerd.run.kata.v2 --name=test ubuntu:24.04 sleep 6h
$ docker exec -it test bash
```

```
$ docker kill test
```

Now our `exec` terminal is hanging (can be resolved by killing the `docker exec` process).

ConnorNelson:
Interestingly, `docker exec -i test bash` under the same setup, can be killed with ctrl+c (but is still incorrect, because the process should automatically die when the container is killed).

PR #919, opened 2025-09-23
SSH: Add workaround for kata docker-exec hang

zrayaz1:
A stopgap fix for #810 

I haven't been able to figure out why kata behaves like this but for now this does fix the observed issue.

### DIFF
--- a/sshd/enter.py
+++ b/sshd/enter.py
@@ -126,6 +126,7 @@ def main():
             runtime = (container.attrs or {}).get("HostConfig",{}).get("Runtime")
             is_kata = runtime == "io.containerd.run.kata.v2"
             if is_kata:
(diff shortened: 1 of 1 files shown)


ConnorNelson:
As a reminder from our discussion yesterday, we want to move this into `sshd/enter.py` with a `.wait()` to detect when the container stops, and sending a SIGSTOP/SIGKILL or whatever to the `docker exec` command. There's a few edge cases around that.

zrayaz1:
I switched the implementation, but it's giving some odd behavior.
Killing the docker exec causes ssh tests to fail, currently they all pass but that's only because I only kill if it's a kata container and we don't test with any of those. It seems to work fine locally I'm not sure why killing the exec causes the tests to fail. 
Another thing is that whenever the docker exec is killed there is be extra white space before the prompt shows up.

zrayaz1:
@ConnorNelson I believe this is good, it works locally

PR #927, opened 2025-09-25
Add API endpoints for use by external applications

TheodorKitzenmaier:
Add API endpoints at `/pwncollege_api/v1/integrations` for use by external applications. The primary purpose of these endpoints is for a **cli program** inside of a challenge container to be able to authenticate. This could be expanded for use outside of the challenge container later, allowing for the ssh-ers to never have to visit the website again.

Something to consider is what capabilities these endpoints should have. A concern is that a malicious dojo/challenge could mess with an unsuspecting dojo user. Someone *other than me* should make decisions on what capabilities a program within the container should have with respect to interacting with pwn.college / CTFd.

Implementation is derived from Yan's work in #767

The `check_authentication` endpoint was added for testing purposes, and can be removed in the future.


TheodorKitzenmaier:
The work has moved to #964

PR #930, opened 2025-09-27
Redesign: Backend changes -- Authentication endpoints + CORS handling

daedlock:


dojo_plugin/api/v1/dojos.py
```diff
@@ -22,12 +23,23 @@
...
         dojos = [
             dict(id=dojo.reference_id,
                  name=dojo.name,
                  description=dojo.description,
-                 official=dojo.official)
-            for dojo in Dojos.viewable(user=get_current_user())
+                 official=dojo.official,
+                 award=dojo.award,
+                 modules=dojo.modules_count,
+                 challenges=dojo.required_challenges_count)
```
ConnorNelson: I think we should probably call these `modules_count` and `challenges_count`.

dojo_plugin/api/v1/dojos.py
```diff
@@ -108,18 +120,30 @@ def get(self, dojo):
...
                  name=module.name,
                  description=module.description,
-                 challenges=[
-                    dict(id=challenge.id,
-                         name=challenge.name,
-                         required=challenge.required,
-                         description=challenge.description)
-                    for challenge in (module.visible_challenges() if not is_dojo_admin
-                                      else module.challenges)
-                 ])
```
ConnorNelson, line 118: This is breaking testcases, and a number of clients expect this data to exist like this. Probably we should just bring it back in its original form, but we can also bring in unified_items.

docker-compose.yml
```diff
@@ -108,6 +108,7 @@ services:
       - MAIL_USERNAME=${MAIL_USERNAME}
       - MAIL_PASSWORD=${MAIL_PASSWORD}
       - MAIL_ADDRESS=${MAIL_ADDRESS}
+      - CORS_ORIGINS=${CORS_ORIGINS}
```
ConnorNelson, line 111: Can we make sure this also ends up in `dojo/dojo-init` as a `define`.

dojo_plugin/api/__init__.py
```diff
@@ -14,9 +16,35 @@
 from .v1.workspace import workspace_namespace
 from .v1.search import search_namespace
 from .v1.test_error import test_error_namespace
+from .v1.user import user_namespace
 
 api = Blueprint("pwncollege_api", __name__)
 
+# Get CORS origin from environment variable
+cors_origin = os.environ.get('CORS_ORIGINS')
```
ConnorNelson: Let's add this to `dojo_plugin/config.py` like our other configuration values, and import it into here from there.

ConnorNelson (changes requested):
Almost there! Thanks :)

daedlock:
> Almost there! Thanks :)

Thanks! all comments were addressed in my last commit

ConnorNelson:
@daedlock It looks like this PR suddenly has way more than the backend stuff.

daedlock:
Oops. My bad i think deploy-local keeps slipping lol. Removing it forever

daedlock:
@ConnorNelson please check. A commit slipped from the workspace changes. Apologies

ConnorNelson:
Awesome stuff, thanks!

PR #932, opened 2025-09-29
CI: Remove vibecheck

ConnorNelson:
Vibecheck has been flaky for too long. It's time has come.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 40
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


zardus:
Why not just make this not be flaky instead of deleting effort?

ConnorNelson:
I haven’t seen it catch something, only lie that things are broken (to the point where I just assume all PRs are going to fail CI… which is a really bad situation to be in). Have you seen it be useful at blocking a PR?

I think LLMs could be useful in code review, but I don’t think it really makes sense as a yolo decider of if the system works or not.

PR #941, opened 2025-10-07
Add latex support with tags

zrayaz1:
A way of having latex be opt in.
Latex would be wrapped in tags like
<latex> </latex>

This might not be the right way to do this but it does work.

### DIFF
--- a/dojo_plugin/utils/__init__.py
+++ b/dojo_plugin/utils/__init__.py
@@ -152,13 +152,21 @@ def render_markdown(s):
         "sub", "sup",
     ]
     markdown_attrs = {
-        "*": ["id"],
+        "*": ["id", "class"],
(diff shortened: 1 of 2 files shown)


Leonabcd123 (changes requested):
Isn't this vulnerable to XSS? We're injecting `latex_content` straight into the div in `replace_latex_section`.

It seems to be currently safe because of `bleach.clean`, but I just want to make sure. This should be fine because we're using `clean_html` for our regex.

zrayaz1:
This allows only `\(`  `\)` and `\[` `\]` to be used as delimiters for latex.
Within markdown pages `\` is treated as an escape so these have to be written as `\\(`  `\\)` and `\\[` `\\]`.
There are probably other edge cases since this renders on top of markdown.

PR #943, opened 2025-10-09
Added practice flag check

Leonabcd123:
I think we should go with #945, so I'll close this once #945 will be merged.


Leonabcd123:
#945 was merged

PR #944, opened 2025-10-10
Site: Refactor index page next steps

hidehiroanto:
Makes it so that the next dojos displayed are the first ones that are not 100% completed according to the required challenge count rather than the total challenge count.


hidehiroanto:
This PR is now mostly moot due to [this commit](https://github.com/pwncollege/dojo/commit/65651325d0c33b8987ec7fa422b3ee5098872d53), it is now just a minor refactoring.

PR #953, opened 2025-10-16
impr: Change emoji validation

Leonabcd123:


zardus:
It looks like tests are timing out on this one? Any ideas?

Leonabcd123:
Looking at the logs, there are some api errors and frontend initialization errors. Maybe those are the problems? It times out even after I removed the tests, so it's either how we download the library or how we use it in `DOJO_SPEC`.

It seems that even the one on master failed. Does that have anything to do with why this one does?

PR #954, opened 2025-10-17
impr: Add target='_blank' to <a> tags

### LINKED ISSUE #903: Suggestion for links found in the Dojo descriptions to open in a new tab
It would be nice if the text links found in the Dojo topic descriptions to use target="_blank" instead of the default target="_self" behavior so it opens in a new tab instead of redirecting from the current pwn.college tab. I think this small change would make the quality of life better as you won't lose your spot on your current page page after clicking on a link. It also saves needing to remember to right click the link to open in a new tab.

Leonabcd123:
Fixes #903


zardus:
@ConnorNelson what do you think?

ConnorNelson:
What about the other return case?

PR #956, opened 2025-10-17
Fixed dojo deletion

### LINKED ISSUE #955: Dojo owners can't delete their own dojos
We're using the `is_admin` function instead of `dojo.is_admin` function in the `delete_dojo` function and when deciding whether to show the delete button to the user, thus only letting pwn.college admins to delete the dojo. This goes against the comment in `delete_dojo`: 
```
# Check if the current user is an admin of the dojo
if not is_admin():
    abort(403)
```

And I believe it's a more logical to also let dojo admins delete their own dojo.

Fix: Change `is_admin` to `dojo.is_admin`

Leonabcd123:
Fixes #955

### DIFF
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -238,8 +238,8 @@ def delete_dojo(dojo):
     if not dojo:
         return {"success": False, "error": "Not Found"}, 404
 
-    # Check if the current user is an admin of the dojo
-    if not dojo.is_admin():
+    # Check if the current user is an admin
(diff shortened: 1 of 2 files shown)


zardus:
The admin restriction was intentional... Dojo deletion has a lot of side-effects (solves deletion, etc) that we're quite terrified of. We were actually considering removing this functionality altogether in favor of some sort of dojo "archiving" that'd just hide the dojo instead...

Leonabcd123:
In that case I'll just change the comment

PR #957, opened 2025-10-20
feat: Add scoreboard filter

### LINKED ISSUE #688: ASU Students Only Leaderboard
Hello,

Professor Yans directed me here to submit a request for improving pwn.college. I was wondering if it is possible to add either a filter tool or a ASU students only section to the leader board. I constantly check the leader boards for each module and can never get an idea as to what rank I place versus my fellow classmates. I know there are a lot of people outside of ASU who use pwn.college, but having a ASU only section would be a nice addition that would help some students become more competitive. I hope that something like this can be implemented!

Leonabcd123:
Fixes #688

**NOTE:**
I'm having some problems with self hosting, so I wasn't able to test any of this code. For that reason (and because I suck), I didn't write the frontend for this (although it should be just a simple select element, and then we can assign its value to `current_filter`).

### DIFF
--- a/dojo_plugin/api/v1/scoreboard.py
+++ b/dojo_plugin/api/v1/scoreboard.py
@@ -141,7 +141,7 @@ def get_scoreboard_page(model, duration=None, page=1, per_page=20):
       user_group = get_group(user.email)
       is_in_filter = group_filter(user_group)
       if is_in_filter:
-        result["standings"].append(standing(item))
+        result["standings"].append(standing(item, user_id))
 
     pages = set(page for page in pagination.iter_pages() if page)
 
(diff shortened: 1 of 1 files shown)


ConnorNelson:
I don't think this is the correct approach. With the exception of showing a fork for asu emails, we do not want special asu-specific code in the dojo.

The correct way to do this sort of thing would be to check if they are a member of the dojo (e.g. DojoMembers or DojoStudents), and probably more specifically if there is an official roster of students for that dojo, only considering those students. But then we'd also need to make sure we don't reveal that roster. I'd want to see all the edge cases of this considered.

Also, unfortunately we cannot merge untested code. For example, your indentation here is incorrect.

Leonabcd123:
So just to make sure before I go any further, we want to only let the official roster of students see the scoreboard that contains only their classmates?

PR #958, opened 2025-10-20
Workspace: Add more tools to nix, refactor .nix files

hidehiroanto:


ConnorNelson:
I think our goal is to intentionally keep the majority of tools on a stable nixpkgs channel.

Is there a specific package that needs to not be?

hidehiroanto:
I was mostly thinking about zed-editor.remote_server, but that can be an overlay, I just didn't see the point of keeping most of the packages on nixos-25.05, given that even nixpkgs-unstable goes through the Hydra CI.

PR #961, opened 2025-10-22
fix(docker_locked) to avoid null JSON on lock expiry

prathamgupta36:
When starting kernel-exploitation challenges, the UI shows “Submission request failed: Cannot read properties of null (reading 'success')”. The frontend expects a JSON object with success, but receives null from the backend and then tries to read result.success.

### To Reproduce the steps are:
  - Add the software-exploitation-dojo.
  - Start a kernel-exploitation challenge.
  - Observe the toast/error banner and console: the POST to /pwncollege_api/v1/docker returns null JSON, followed by the JS error.

### Root Cause
  - The docker_locked decorator uses a Redis lock with timeout=20.
  - If the wrapped function takes longer than 20s, the redis client (older redis-py) raises LockNotOwnedError on exit.
  - The decorator catches LockNotOwnedError and does pass, causing the wrapper to return None.
  - Flask serializes None as JSON null.
  - Frontend code in dojo/dojo_theme/static/js/dojo/challenges.js then does if (result.success) { ... } and crashes on null.

### Affected Code - [`dojo/dojo_plugin/api/v1/docker.py`](https://github.com/pwncollege/dojo/blob/master/dojo_plugin/api/v1/docker.py#L338)
  - Backend lock: [dojo/dojo_plugin/api/v1/docker.py:338](https://github.com/pwncollege/dojo/blob/master/dojo_plugin/api/v1/docker.py#L338) (function docker_locked)
  - Frontend usage: dojo/dojo_theme/static/js/dojo/challenges.js:243 (reads result.success after response.json()), catch prints “Submission request failed: …” on exception dojo/dojo_theme/static/js/
  dojo/challenges.js:286.

### Proposed Changes
  - Replaced context-managed lock with explicit acquire/release to ensure a well-formed response is always returned.
  - If lock cannot be acquired immediately, return a JSON error instead of falling through.
  - Always attempt to release the lock in a finally block and ignore release errors so the wrapped function’s response is preserved.

[dojo/dojo_plugin/api/v1/docker.py:338](https://github.com/pwncollege/dojo/blob/master/dojo_plugin/api/v1/docker.py#L338)

Removed Things:
      - Context manager pattern that could raise LockNotOwnedError and swallow it, returning None.
  - Added:
      - Manual lock = redis_client.lock(..., timeout=20), acquire(blocking=False).
      - Early return: {"success": false, "error": "Already starting a challenge; try again in 20 seconds."} when not acquired.
      - finally with lock.release() guarded by except redis.exceptions.LockNotOwnedError and except redis.exceptions.LockError to avoid breaking the response.

### Reason for changes:
  - Minimal, targeted change: preserves existing semantics while guaranteeing a valid JSON response.
  - Avoids null responses which break multiple frontend code paths.
  - Keeps the same 20s TTL and concurrency behavior; only fixes the “return None on lock release error” bug.

### Testing - `I have not tested this due my MCP dojo dev`
  - Before: POST /pwncollege_api/v1/docker sometimes returns null when container start exceeds 20s; frontend throws.
  - After: The same request returns either:
      - {"success": true} when started; or
      - {"success": false, "error": "Already starting a challenge; try again in 20 seconds."} when lock busy; or
      - Proper error JSON when the wrapped function returns an error.
  - Frontend no longer throws “Cannot read properties of null …” and displays info/warn banners as intended.

### DIFF
--- a/dojo_plugin/api/v1/docker.py
+++ b/dojo_plugin/api/v1/docker.py
@@ -339,19 +339,29 @@ def docker_locked(func):
     def wrapper(*args, **kwargs):
         user = get_current_user()
         redis_client = redis.from_url(current_app.config["REDIS_URL"])
+        lock = redis_client.lock(
+            f"user.{user.id}.docker.lock", blocking_timeout=0, timeout=20
+        )
         try:
-            # Super annoying that this will throw a LockNotOwned
-            # exception if the func takes longer than the timeout The
-            # real fix is raise_on_release_error=False to .lock (added
-            # https://github.com/redis/redis-py/issues/3532), however
-            # we're using an old version of the redis client. So for
-            # now we just catch LockNotOwnedError
-            with redis_client.lock(f"user.{user.id}.docker.lock", blocking_timeout=0, timeout=20):
-                return func(*args, **kwargs)
-        except redis.exceptions.LockNotOwnedError:
-            pass
+            acquired = lock.acquire(blocking=False)
         except redis.exceptions.LockError:
-            return {"success": False, "error": "Already starting a challenge; try again in 20 seconds."}
+            acquired = False
+
+        if not acquired:
+            return {
+                "success": False,
+                "error": "Already starting a challenge; try again in 20 seconds.",
+            }
+
+        try:
+            return func(*args, **kwargs)
+        finally:
+            try:
+                lock.release()
+            except redis.exceptions.LockNotOwnedError:
+                pass
+            except redis.exceptions.LockError:
+                pass
     return wrapper
 
 


ConnorNelson:
@prathamgupta36 Interesting, thanks for figuring out the root cause!

I think I'm going to resolve it with #965, addressing the massive comment that was in the locking code. Does that seem reasonable to you?

ConnorNelson:
I'd guess the issue happening specifically with the kernel security challenges is copying the vmlinux in. It's a little worrying though that it can't transfer 400mb fast enough, that's a little surprising. Realistically though, the solution is to put the vmlinux into challenge images so it doesn't need to transfer on start.

prathamgupta36:
The code in #965 seems cleaner and doesn't acquire and release like my code does. Seems like a better fix. I did find it weird that it was taking long enough to transfer to trigger the error.

PR #964, opened 2025-10-22
Workspace: Add `dojo` Command-Line Application

TheodorKitzenmaier:
# Overview
Adds a command line application, `dojo`, which allows for limited interaction with a custom set of integration APIs. The application authenticates using the `DOJO_AUTH_TOKEN` environment variable, which has been changed to be a set of *signed values* tying the token to the user and their current challenge.

In order to allow for communication between the application and the dojo, **challenge containers** have been given network access to the **nginx container**. In theory the iptables configuration in `dojo/dojo-init` should ensure that this is the *only* other container that the challenge containers can access, but someone more familiar with iptables should sanity check this.

# Integrations
Added a new namespace to the pwn.college api, `integrations`. Endpoints within the integrations namespace expect requests to use the ~~`auth_token`~~ `AuthToken` header to provide the current container's authentication token.

# Command
## whoami
A simple command which prints out information about the current user. Prints the user's name and id.

Invoked by calling `dojo whoami` in a terminal.

### DIFF
--- a/dojo_plugin/api/v1/docker.py
+++ b/dojo_plugin/api/v1/docker.py
@@ -11,6 +11,7 @@ import docker.errors
 import docker.types
 import redis
 from flask import abort, request, current_app
+from itsdangerous.url_safe import URLSafeTimedSerializer
 from flask_restx import Namespace, Resource
 from CTFd.cache import cache
 from CTFd.models import Users, Solves
@@ -24,7 +25,6 @@ from ...utils import (
     container_name,
     lookup_workspace_token,
     resolved_tar,
-    serialize_user_container,
     serialize_user_flag,
     user_docker_client,
     user_node,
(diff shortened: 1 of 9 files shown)


ConnorNelson:
- Let's change dojo-cli so its not inlined in a nix file, its in its own file (with nice syntax highlighting, etc).
- Let's get rid of `integration.py`, instead lets have a wrapper that we apply to other end points. For example, `@authed_only` is the standard CTFd wrapper, you added `@authed`. Lets rename `@authed` to `@authed_only_cli`, which can handle the `@authed_only` mechanism of authentication AND also the container-based auth-token. Then for the endpoints we want to make available to cli, we'll swap the `@authed_only` to a `@authed_only_cli`.

TODO for me is to further investigate the networking.

ConnorNelson:
@codex Please fix any python styling issues, e.g. `userID, challengeID`.

ConnorNelson:
@codex Please fix any python styling issues, e.g. `userID, challengeID`.

zardus:
Wooo!

PR #966, opened 2025-10-22
Frontend: Add Bun-based Next.js service for future deployment

daedlock:
New envs:
```
FRONTEND_HOST=future.pwn.college
```

Set env:
```
CORS_ORIGINS=https://future.pwn.college
```

### DIFF
--- a/nginx/conf.d/server-http-future.dev.conf
+++ b/nginx/conf.d/server-http-future.dev.conf
@@ -1,5 +1,5 @@
 server {
-    include /etc/nginx/conf.d/frontend-host.conf;
+    include /etc/nginx/conf.d/future-dojo-host.conf;
     listen 80;
 
     location / {


ConnorNelson:
Thanks! This is, of course, super exciting and awesome!

Some minor high level feedback that we may want to address in a future PR: it might be nice to not require the NEXT_PUBLIC_* prefix variables at build time. Can the frontend somehow just get this from window.location, and backend from host header? This would allow us to have a single build artifact that works for any deployment, and also allow ssh forwarding (with different ports) to work seamlessly. For example, `ssh -L localhost:8080:172.17.0.2:80 ...`, and then accessing it on `future.localhost.pwn.college:8080`. As it currently stands, being on a different port breaks things (the first symptom I observed is the login page).

nuggetpilot:
NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

PR #967, opened 2025-10-23
CI: Make coverage report upload in single-node mode only

maxtrap:
Fixed a small oversight in my original PR where coverage reports are attempted to be uploaded in multi-node mode as opposed to being skipped.

It should be skipped because the report isn't generated in multi-node.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,7 @@ jobs:
           fi
 
       - name: Upload coverage reports to Codecov
+        if: matrix.mode == 'singlenode'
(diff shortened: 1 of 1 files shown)


maxtrap:
Any idea why multinode tests are failing to start? It's only failing sometimes.

supercoolspy:
> Any idea why multinode tests are failing to start? It's only failing sometimes.

I fixed it in #721 (among other things) you need to restart nginx after restarting sshd and ctfd since it changes the route to their container which nginx only resolves once

ConnorNelson:
Thanks!

PR #971, opened 2025-10-24
fix: Fix links to dojos in feed page

### LINKED ISSUE #970: Links to dojos in feed page don't work
All dojo links throw a 404 error, because we're trying to access /dojos/{DOJO_ID} instead of just /{DOJO_ID}.

Leonabcd123:


Leonabcd123:
Why do we even have this part of the code twice?

PR #988, opened 2025-11-12
Add foldered "medal" awards

TheodorKitzenmaier:
Adds support for a new type of awards, "medals".

Medals can be granted for first, second or third place, and are tied to the name of an event on the dojo. Medals are granted by admins of any dojo which has the `grant_awards` permission. There is support for granting, revoking, and pruning event awards.

They are displayed alongside normal emoji awards, however if a user has multiple medals of the same tier, they will be *foldered*. In practice, this means that only a single medal will show, however the description shown when hovering over the award will show all events that tier of medal was won in.

This feature is part of my Capstone project at ASU to support competitive events on the dojo infrastructure.


TheodorKitzenmaier:
Redis cache has destroyed me.

PR #989, opened 2025-11-13
Feature: Ability to link files opening in new page

prathamgupta36:
## Dojo Asset Links Support

Dojo owners can now embed links inside dojo, module, and challenge descriptions that point to files stored in their dojo repository. Visiting a URL of the form `/dojo_id/<asset-path>` renders the referenced artifact directly in the platform:

- `/{dojo_id}/{filename}` resolves dojo-level files.
- `/{dojo_id}/{module_id}/{filename}` resolves module-level files.
- `/{dojo_id}/{module_id}/{challenge_id}/{filename}` resolves challenge-level files.

Markdown renders as HTML, PDFs stream in the browser, and LaTeX (or other plain text assets) display in a dedicated template. Requests are constrained to the owning dojo repo and reject ambiguous matches, preventing traversal outside the allowed tree.

## Motivation
For the core material to be added for cryptography dojo, having long pdfs have their own dedicated page is a necessity. Previously, descriptions could only link to external hosting or expose bare markdown. This feature keeps assets in the dojo repository, preserving version control, discoverability, and offline deployability while giving students a single, predictable URL structure for supplementary content.

## Implementation Details
- **Routing**: `dojo_plugin/pages/dojo.py` now inspects deep paths and dispatches file requests after validating path segments. Files resolve relative to the dojo tree only when the match is unique. Safe extensions are rendered appropriately; everything else streams via `send_file`.
- **LaTeX/Plain Text Template**: `dojo_theme/templates/text_file.html` renders textual assets with dojo/module context headers.

## Files Changed
- `dojo_plugin/pages/dojo.py`: Added file-resolution helpers, sanitized segment checking, and expanded `view_dojo_path` routing.
- `dojo_theme/templates/text_file.html`: New template for text-based assets.

`UNDER CONSTRUCTION`

### DIFF
--- a/dojo_plugin/pages/dojo.py
+++ b/dojo_plugin/pages/dojo.py
@@ -1,9 +1,11 @@
 import collections
+import os
 import subprocess
 import traceback
 import datetime
 import sys
 import re
+from pathlib import Path
 
 from flask import Blueprint, render_template, abort, send_file, redirect, url_for, Response, stream_with_context, request, g
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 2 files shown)


ConnorNelson:
I think the `pages` attribute in the dojo.yml already supports this use-case. Please let me know if that's not accurate.

PR #995, opened 2025-11-30
Site: Restrict access to other users' workspaces

### LINKED ISSUE #984: Users can start services on other users' machines
In the `pwncollege_api/v1/workspace` endpoint, we get the service and user id from the user:

```
user_id = request.args.get("user")
password = request.args.get("password")
service = request.args.get("service")

if user_id and not password and not is_admin():
    abort(403)

 user = get_current_user() if not user_id else Users.query.filter_by(id=int(user_id)).first_or_404()
```

Afterwards, we call `start_on_demand_service` with that service and user:

```
if start_on_demand_service(user, service) is False:
    return {"success": False, "active": True, "error": f"Failed to start service {service}"}
```

In `start_on_demand_service`, we check whether the service is terminal, code or desktop, and if it is one of them, we start it:

```
on_demand_services = { "terminal", "code", "desktop"}

def start_on_demand_service(user, service_name):
    if service_name not in on_demand_services:
        return None
    try:
        exec_run(
            f"/run/current-system/sw/bin/timeout -k 10 30 /run/current-system/sw/bin/dojo-{service_name}",
            workspace_user="hacker",
            user_id=user.id,
            assert_success=True,
            log=True,
        )
```

This allows anyone to start the terminal and code services on another users' machine. Note that we cannot start the desktop service because we verify the user when the service is desktop in the `pwncollege_api/v1/workspace` endpoint:

```
if service == "desktop":
    interact_password = container_password(container, "desktop", "interact")
    view_password = container_password(container, "desktop", "view")

    if user_id and password:
        if not hmac.compare_digest(password, interact_password) and not hmac.compare_digest(password, view_password):
            abort(403)
        password = password[:8]
    else:
        password = interact_password[:8]
```

supercoolspy:
Needs to be avoided now since an unauthenticated url would be returned in the api
Fixes #984

### DIFF
--- a/dojo_plugin/api/v1/workspace.py
+++ b/dojo_plugin/api/v1/workspace.py
@@ -68,6 +68,10 @@ class view_desktop(Resource):
 
         signature = base64.urlsafe_b64encode(digest).decode()
 
+        if not service == "desktop":
+            if user_id and not is_admin():
+                abort(403)
+
         if service:
             if service == "desktop":
                 interact_password = container_password(container, "desktop", "interact")
@@ -95,9 +99,6 @@ class view_desktop(Resource):
                 iframe_src = forward_workspace(service=service_param, service_path="vnc.html", signature=signature, message=message, **vnc_params)
 
             elif service == "desktop-windows":
-                if user_id and not is_admin():
-                    abort(403)
-
                 service_param = "~".join(("desktop-windows", str(user.id), container_password(container, "desktop-windows")))
                 vnc_params = {
                     "autoconnect": 1,


ConnorNelson:
Good catch! Can we change it though so that:
1. the user is allowed to get a url
2. a dojo admin is allowed to get a url for another user (e.g. `dojo.is_admin` rather than `is_admin`)

And let's do this for all cases. We can later think about a user sharing the actual direct link that goes straight to their workspace.

supercoolspy:
>     1. the user is allowed to get a url

Do you mean by setting the user id field to their own user? Otherwise I believe that `user_id` is only set when you aren't trying to get a url for your account.

ConnorNelson:
I meant the normal base case (which is already covered).

PR #996, opened 2025-12-02
Site: Add setuptools dependancy

zrayaz1:
CTFd is looking for distutils which appears to no longer be apart of base python.


ConnorNelson:
Sweet, thanks!

PR #1004, opened 2025-12-19
Site: Fix clipboard copy and paste

### LINKED ISSUE #1003: Copy/Paste in workspace is (programmatically) broken
Now that we're on different domains, the clipboard api doesn't have permission, at least on Google Chrome.

This can be reproduced by trying to paste something into vscode terminal, or trying to hit the copy flag button at the end of a lecture.

Copy/paste into vscode editor view works fine, and copy from vscode works fine.

I think that ordinarily it would all work, but chrome would give a popup asking if you want to give clipboard permissions. Now, the popup never even comes (maybe because its a cross-site iframe?).

If you manually navigate to the iframed url, and perform the restricted actions, chrome prompts for permission, but we need a smoother workflow than that ideally.

---

I tried adding this to the iframe `allow="clipboard-read; clipboard-write self https://asu-2-workspace.pwn.college"`, based on https://hellodevworldblog.medium.com/error-when-copying-to-clipboard-from-iframe-dab0b5c5f310, but no luck. I would imagine that maybe I just have the syntax wrong. It seems like there might also be headers that can influence all of this stuff.

ConnorNelson:
### Motivation
- Resolve issue where copy/paste is blocked inside the embedded workspace (issue #1003).
- Browsers require explicit iframe permission hints to allow clipboard access for framed content, especially across origins.

### Description
- Grant clipboard permissions to workspace iframes by adding `allow="clipboard-read *; clipboard-write *"` in three places:
  - `dojo_theme/templates/module.html` — static iframe in the module template.
  - `dojo_theme/static/js/dojo/challenges.js` — dynamically created challenge iframe inserted into the DOM.
  - `frontend/src/components/workspace/WorkspaceService.tsx` — React/TSX workspace iframe component used by the frontend.
- This is a minimal change that surfaces the appropriate permissions hint to browsers so clipboard read/write requests from the framed workspace can be granted.

### Testing
- Automated tests: none were run for this change.
- (No automated test suite executed as part of this patch.)

-----
Resolves #1003.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_694574b3f6f08333a4bebf1310c328b3)

### DIFF
--- a/dojo_theme/static/js/dojo/challenges.js
+++ b/dojo_theme/static/js/dojo/challenges.js
@@ -263,7 +263,7 @@ function startChallenge(event) {
         $(".challenge-workspace").addClass("challenge-hidden");
         $(".iframe-wrapper").html("");
         if (result.success) {
-            item.find(".iframe-wrapper").html("<iframe id=\"workspace-iframe\" class=\"challenge-iframe\" src=\"\"></iframe>");
+            item.find(".iframe-wrapper").html("<iframe id=\"workspace-iframe\" class=\"challenge-iframe\" src=\"\" allow=\"clipboard-read *; clipboard-write *\"></iframe>");
             loadWorkspace();
             item.find(".challenge-init").addClass("challenge-hidden");
             item.find(".challenge-workspace").removeClass("challenge-hidden");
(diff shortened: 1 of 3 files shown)


supercoolspy:
Fix looks good!

Issue #951, opened 2025-10-15
Allow dojo admins to change the repository the dojo is based on


Leonabcd123:
Currently, there's no way to change `dojo.repository`. There should be a way to change the repository a dojo is based on, for cases such as moving the repository to a different one. An example of a dojo that would benefit from this change would be the arm dojo. `dojo.repository` is currently `robwaz/arm-dojo`, although it has been relocated to `pwncollege/arm-dojo`.

Also, is this considered information disclosure? Even when the dojo is private, it's still possible to get the name, owner, and branch name of the repository (via the edit description button). This information disclosure issue will be fixed when #942 will be fixed.

PR #1009, opened 2025-12-22
Site: Update dojo from JSON spec

ConnorNelson:
## Summary
- add admin-only API endpoint to update a dojo from JSON spec
- ignore git repo by applying spec directly

## Testing
- not run (not requested)


ConnorNelson:
Once #1006 is in, we can add a test to make sure this new endpoint works, and also get rid of the test endpoint for deleting recent solves, and instead just update the dojo to have an additional challenge.

supercoolspy:
Would also solve #951

PR #1016, opened 2026-01-05
Infrastructure: Use ephemeral /data in deploy.sh

ConnorNelson:
## Summary
- simplify local deploys by relying on ephemeral `/data` inside the container

## Testing
- not run (per request)


zardus:
Please let's not do this. We've done something like this in the past, but the result is /data of failed testcases is, by default, lost. This makes deving a lot harder, which is why I made persistent /data the default.

PR #1036, opened 2026-01-22
Workspace: Expand `dojo` Command Line Features

TheodorKitzenmaier:
Implements additional features of the `dojo` application. Planned features for inclusion in this pr are:
- Flag submission
- Restarting the current challenge
- Starting a new challenge


hidehiroanto:
parse_dojo_path makes no sense, you can't have module or challenge set to None, just do `challenge` or `/dojo/module/challenge` using regex

TheodorKitzenmaier:
For this PR it doesn't really have a purpose, I implemented it for reuse as part of the the `dojo list` command where you might be listing the modules within a dojo, so module and challenge would be `None` in that case. It might still be cleaner to handle that as its own case.

PR #1093, opened 2026-05-12
SSH: account creation and key upload 

zrayaz1:
wip


zardus:
Wow, this is so much more elegant than the other attempt. What's left before we merge?

zrayaz1:
> Wow, this is so much more elegant than the other attempt. What's left before we merge?

Right now I'm having it still require users to give a username and email,  and just setting their password to a random sequence so if they wanted to later use the website it would require going through the password reset process. I'm not sure if that's the best way to do it or if we would prefer something else. 

The other part of this is an easier way of adding ssh keys, when doing ssh with a key that is not already affiliated with an account the user is prompted to either create an account or add that key to an existing account and then it generates a link which when followed would add the key. Also not sure if this is a good idea and might have some funny concerns about getting someone to click a link.

zardus:
Is this a stale PR?

zrayaz1:
> Is this a stale PR?

Not anymore! Still needs review. I stripped it down to just what I think is needed for account creation to work. 
Also looking for ui feedback for this

PR #1095, opened 2026-05-25
Workspace: Fix rust in nix

hidehiroanto:
Apparently rustup is not the way to go when it comes to installing Rust using Nix
https://nixos.wiki/wiki/Rust


supercoolspy (approved):
Tested locally w/ NixOS and packages worked

PR #1100, opened 2026-07-07
SSH/Workspace: Move tui into dojo-cli 

zrayaz1:
Moves the tui into dojo-cli as part of nix. 

ssh now subprocesses `dojo tui`

Fixes local development stuff with DOJO_HOST, and fixes terminal state corruption on container re-entry.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -264,6 +264,7 @@ services:
       - /data/ssh_host_keys:/etc/ssh:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
       - /data/mac:/var/mac:ro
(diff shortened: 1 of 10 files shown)


zrayaz1:
@ConnorNelson Is there a better way to do this? I wanted to keep it in nix and just mount nix in, but I forgot about multinode. I do think its better if there is a way to keep it all in nix as it moves the dependencies out of the sshd.

Currently this one is just mounting `/opt/pwn.college/workspace/core/tui.py:/opt/sshd/tui.py:ro` 
I would prefer to just subprocess `dojo tui`  in enter.py

PR #1106, opened 2026-07-14
Index/Register: simplify the front page

zardus:
## Summary

- remove the writeup rule and unenforced registration commitment gate
- remove the index-page rules gate and the requested informational sections
- split the introductory text into shorter paragraphs without rewriting its meaning
- keep the bug-bounty ground rule

## Testing

- update the registration browser test to submit directly


sudorm-fmeee:
ddddddddddd

PR #1110, opened 2026-07-24
Detach converted imported challenges

### LINKED ISSUE #1037: Converting an imported challenge to a non-imported challenge is weird
Haven't dug in too much, but it seems that making an imported challenge into its own thing requires first deleting it.j

zardus:
Fixes #1037

## Summary

- exclude imported challenge associations when looking for a native challenge to reuse
- removing `import` now creates a destination-owned challenge instead of mutating the source challenge

## Testing

- convert an imported challenge without deleting it first
- verify the source is unchanged, the destination gets its own flag and identity, and repeating the update is stable


zardus:
@ConnorNelson thoughts on the way this guy is checking for imports? It's the only way that doesn't FP on previously-transferred chals without database changes. Probably okay for now?

PR #1113, opened 2026-07-24
Key module scoreboards by module ID

### LINKED ISSUE #1097: scoreboard with full solves on a fresh module
This seems not right

Image

zardus:
Fixes #1097

## Summary

- key module scoreboard and crew caches by the stable module ID instead of its mutable list index
- add an explicit `id` discriminator so numeric module IDs cannot collide with legacy index keys
- change only the four module scoreboard readers and writers involved in the reported bug

## Testing

- populate a module scoreboard, replace that module at the same index, and verify the replacement starts empty
- solve the replacement once and verify its score is exactly one


zardus:
@adamdoupe do you see any issues with this? It sets module ID instead of index for the cache key, to avoid the issue where inserting a module midway through a dojo makes things really weird. Other than having to regenerate caches on start, it should be good, right?

PR #1154, opened 2026-09-09
Fix hidden users' own progress bars within a dojo

mikenex10:
If you set your account's "Visibility" setting to Hidden (so you don't show up on the scoreboard), the progress bars inside a dojo get stuck at 0%, even after you've actually solved challenges. This only affects the view inside a dojo, not on the main dojos listing page.

The per-module progress bar was using the same filter used to hide a user's solves from the public scoreboard. That filter makes sense for rankings, but it was being applied when calculating a user's own progress too, so a hidden user's own solves were being excluded from their own progress bar.

This PR makes "Hidden" setting only affect scoreboard visibility (as I think it's intended) and no longer hides your own progress from yourself.

### DIFF
--- a/dojo_plugin/models/__init__.py
+++ b/dojo_plugin/models/__init__.py
@@ -525,7 +525,6 @@ class DojoModules(db.Model):
     def visible_solves(self, **kwargs):
         return self.solves(ignore_visibility=True, **kwargs).filter(
             DojoChallenges.visible(),
-            ~Users.hidden,
         )
 
     @hybrid_method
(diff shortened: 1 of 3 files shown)


ConnorNelson:
Thanks for the fix! I merged in `master` following #1157 and updated the progress call to pass `include_hidden_users=True`. This keeps hidden users excluded by default while allowing them to see their own progress. I also added a regression test covering the hidden user, another user, and an anonymous viewer.

PR #1166, opened 2026-09-13
Upgrade Kata Containers to 4.1.0

ConnorNelson:
Upgrade Kata Containers from 3.19.1 to 4.1.0 and use the Rust runtime. Update archive extraction and enable debug info to preserve BTF in the guest kernel.

Mount the Docker node's resolver configuration into Kata workspaces so external DNS works with both `bridge` and `workspace_net` attached. Add DNS and HTTPS tests for standard and practice modes, including multi-node coverage.

Tested locally:

- Docker image and guest kernel build.
- Full suite: 714 passed, 15 skipped, 1 existing expected failure.
- Multi-node suite: 32 passed.
- Final DNS and HTTPS tests: 2 passed.

The existing expected failure is `test_workspace_urls_stop_serving_once_the_container_is_gone`, which checks that a terminal URL returns 404 after its container is deleted.


zrayaz1:
I believe this will fix the ssh hang issue. The issue appears to come from the go runtime, I couldn't reproduce it in the rust runtime. When this gets merged we should be able to revert #919 for #810

PR #1174, opened 2026-09-22
Remove CTFd: run the dojo as a standalone Flask app

zardus:
## Summary

The dojo has been a CTFd plugin, but we use less and less of CTFd. This PR removes CTFd entirely and runs `dojo_plugin` as a standalone Flask application (`dojo_plugin.create_app()`), with narrow reimplementations of only the parts the dojo actually used.

**The database is fully compatible: no migration.** An existing production database works unchanged, and stays usable by the previous CTFd-based code at any time (rollback safety). Fresh installs produce the same schema and seed rows as before. Live sessions survive the switch. The only intentional functional removal is `/admin`.

## What was reimplemented (`dojo_plugin/core/`)

- `models.py`: minimal ORM models for the CTFd-owned tables the dojo uses (`users`, `challenges`, `flags`, `submissions`/solves/fails, `awards`, `tokens`, `tracking`, `config`, `pages`), declared DDL-equivalent to the live schema (column order, constraints, the `users_email_lower_key` index and case-insensitive email comparator, the `bcrypt_sha256` password hook). `legacy_tables.py` declares the unused CTFd tables as bare tables so fresh installs create all 36 tables and the `alembic_version` row.
- `sessions.py`, `signing.py`, `security.py`, `passwords.py`: cookie name, sid signing, redis key layout and payload encoding are byte-compatible with what CTFd wrote, so nobody is logged out at cutover; flags, workspace/ssh tokens, email links and API tokens keep verifying.
- `identity.py`, `decorators.py`, `hooks.py`: `get_current_user`/`is_admin`/`get_ip`, `authed_only`/`admins_only`/`ratelimit` (same redis keys and 429 body), and the per-request hooks (tracking, banned, `Token`/`Bearer` authorization handling, CSRF) in the same order and with the same semantics.
- `siteconfig.py`: `get_config`/`set_config` over the `config` table with CTFd's coercion and memoization. Reads stay bug-compatible with the duplicate-key rows production has; writes no longer create duplicates.
- `mail.py`, `markdown.py`, `forms.py`, `validators.py`, `countries.py`, `theme.py`, `errors.py`, `auditlog.py`, `dates.py`, `challenges.py`: the remaining helpers (mail now carries a `Date:` header, which the old monkeypatch intended but never achieved).
- `pages/auth.py` and `pages/views.py`: the HTML login/register/logout/confirm/reset-password flows, `/settings`, `/themes/...` static serving, `/challenges` -> `/dojos`, `/events`, `/healthcheck`, all under the same endpoint names.
- `ctfd_api/`: the `/api/v1` endpoints the theme JS and clients still call (`challenges/attempt`, `users`, `users/me`, `users/<id>`, awards, `tokens`) with identical JSON envelopes, validation messages and error-body semantics.
- `bootstrap.py`: schema creation and seed rows run once from the container entrypoint (`python -m dojo_plugin.bootstrap`), so the old multi-process boot race that produced duplicate config rows is gone. Workers are explicit entry points (`python -m dojo_plugin.worker`, `python -m dojo_plugin.worker.image_pulls`) and wait for the schema.

## Theme

The CTFd core theme's production bundles (`vendor.bundle.js`, `core.js`, `helpers.js`, `pages/main.js`, `pages/settings.js`, CSS, fonts, sounds) and the nine core templates the dojo rendered through CTFd's theme fallback are vendored into `dojo_theme/` (Apache-2.0 notice included). The `.dev/.min` URL mangling, the symlink farm and the `?d=<run_id>` cache-buster are gone; every js/css asset now gets a content-hash `?v=` and `Cache-Control: max-age=3600`.

## Ops

- `ctfd/` now holds a small Dockerfile, `entrypoint.sh` and the trimmed requirements (same pins as before for everything that survives). The outer Dockerfile no longer fetches or patches CTFd.
- The compose service is still named `ctfd` and the package is still `dojo_plugin`; renaming those identifiers is a follow-up to keep this diff reviewable.
- `dojo sync` is gone (the code is bind-mounted); `dojo flask`, `dojo load-dojo`, `dojo db` work as before.

## Removed on purpose

`/admin/*` (including `/admin/dojos` and the navbar link), custom user fields (tables stay), the `pages` mechanism, teams, notifications/SSE, uploads, the update-check telemetry, MajorLeagueCyber OAuth, Flask-Migrate, marshmallow and the other CTFd-only dependencies.

## Tests

Test edits are limited to what the removal implies: import paths, the `/admin` and custom-fields tests, the asset URL regex, a hard-coded `/opt/CTFd` path, a helper that read the dead `solves` table, and the `test_background_stats` unit tests, which used to run through the interactive shell (which always exited 0) and now run for real. Two of those tests had latent defects that the old vacuous runner hid (a MagicMock used inside a SQL expression, a string passed for an integer id); both are fixed.

## Verification done locally

- Fresh install: normalized `pg_dump --schema-only` is identical to a CTFd-produced database; 18 config rows in the same order, no duplicates, `alembic_version` = `9e6f6578ca84`, hidden `admin` user.
- In-place upgrade of a running CTFd-era deployment (database and redis written by the old code): schema byte-identical afterwards, sessions minted by the old code still authenticate, existing users log in, flag submission through `/api/v1/challenges/attempt` and the dojo API works, workers start.
- Rollback: the previous CTFd-based code boots on a database bootstrapped and written by this branch, its migrations are a no-op, and it logs in and serves its pages including `/admin`.
- Flag/token/email-link signing, the HTML auth flows, audit log lines, the hook order and the error-body semantics on both API prefixes were probed against a live deployment.

## Things worth checking on production before deploying

- `SELECT type, count(*) FROM challenges GROUP BY 1` (and the same for `flags`): only the `dojo` type is implemented; any pre-dojo CTFd challenge rows of another type would 500 on `/api/v1/challenges/attempt`.
- `/api/v1/users/me` `score`/`place` are still computed the CTFd way; if production has no valued challenges or awards these can become constants.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zardus:
CI note: the first semantic-tier run timed out after the workspace tests wedged the homefs volume plugin (a pre-existing race that also reproduces on master; details and fix in #1175). The tier has been re-run; if it wins the race it is green, otherwise this branch will be rebased onto #1175.

